PR #7, opened 2019-12-09
Update ReadMe

K-Bains:


README.md
```diff
@@ -47,6 +49,18 @@ Firstly, you'll need to enable some features. Expand 'Configuration' and select
 
 ![Enable Features](/Images/GraphSyncFeature.PNG)
 
+#### Create Content Item
```
lazcool: This copy would be better in the user guide (linked to below), rather than in the developer environment setup.

README.md
```diff
@@ -22,7 +22,9 @@ Click 'Add Graph', then 'Create a Local Graph'. Enter a graph name, set the pass
 
 That's all you need to do for syncing within the editor to work. To perform interactive queries against the graph, once the graph is Active, click 'Manage' and then on the next page, click 'Open Browser'. If you're unfamiliar with the browser or Neo4j in general, check out the [docs](https://neo4j.com/developer/neo4j-browser/).
 
-### Create Content Database
+To install the ESCO TTL file into Neo4J there are a number of steps that need to be followed and executed, for more information on how to do this view [Neo4J Setup](https://skillsfundingagency.atlassian.net/wiki/spaces/DFC/pages/1491501074/SPIKE+RDF+hosting+-+Neo4j+cluster) on Confluence
```
lazcool: Worth bearing in mind that this readme is public, whereas the Confluence page isn't.

Perhaps mention the page linked to isn't public, or add a readme with the instruction to load the ESCO data.

Might be worth mentioning that loading ESCO data is optional and only required for the NCS use case?

PR #15, opened 2020-01-28
Ncsd 2303 automate deployment

NickGraham101:


MonoSample/program.cs
```diff
@@ -0,0 +1,73 @@
+using System;
```
lazcool: Do we need this MonoSample project?

SimonWren: removed

PR #30, opened 2020-02-14
Bag Part Sync

lazcool:


tomtollafield:
Reviewed and looks okay to me. Integral parts to the sync process have changed so would be good for @sturatcliffe to have a look as well as I am not yet 100% with the inner workings.

PR #39, opened 2020-02-26
Delete node/relationships from neo when content item is deleted

sturatcliffe:
… No restrictions or failure handling yet.


sturatcliffe:
Can no longer delete a content item / node when it has incoming relationships. If the neo sync fails for whatever reason, the OC transaction is rolled back to remain consistent.

However, there is no easy way of preventing the "Content item was deleted" success message from displaying.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/GraphSyncer.cs
```diff
@@ -83,6 +90,24 @@ public async Task SyncToGraph(ContentItem contentItem)
             await SyncComponentsToGraph(graphSyncPartContent);
         }
 
+        public async Task DeleteFromGraph(ContentItem contentItem)
+        {
+            _deleteNodeCommand.ContentType = contentItem.ContentType;
+            _deleteNodeCommand.Uri = contentItem.Content.GraphSyncPart.Text;
```
lazcool: We should silently handle the case where the content item being deleted hasn't got a graph sync part, i.e. do nothing.
We don't whitelist content types in the triggering event and use the presence of a graph sync part as a flag to sync the content item.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/GraphSyncer.cs
```diff
@@ -83,6 +90,24 @@ public async Task SyncToGraph(ContentItem contentItem)
             await SyncComponentsToGraph(graphSyncPartContent);
         }
 
+        public async Task DeleteFromGraph(ContentItem contentItem)
```
lazcool: We should move this into its own delete specific graph sync class (and probably rename this class to something akin to UpsertGraphSyncer or better 😄). They have quite different dependencies.

PR #41, opened 2020-02-27
DST-136 Better Sync Validation

lazcool:


DFC.ServiceTaxonomy.Neo4j/Services/NeoGraphDatabase.cs
```diff
@@ -1,42 +1,51 @@
...
+                o => o.WithLogger(neoLogger));
         }
 
-        public async Task<List<T>> RunReadQuery<T>(Query query, Func<IRecord, T> operation)
+        public async Task<List<T>> RunReadQuery<T>(IQuery<T> query)
         {
             IAsyncSession session = _driver.AsyncSession();
             try
             {
                 return await session.ReadTransactionAsync(async tx =>
```
tomtollafield: If the read or write transactions throw an exception, is it worth logging in the finally clauses?

lazcool: Any exceptions thrown within the transaction, get logged by Neo/NeoLogger, e.g. if a command validation fails, we see this in the logfile...
```
DFC.ServiceTaxonomy.Neo4j.Log.NeoLogger|ERROR|[null] DFC.ServiceTaxonomy.Neo4j.Exceptions.CommandValidationException: Expecting 1 node to be deleted, but 0 were actually deleted.
```

PR #43, opened 2020-02-27
- Add Created Date and Last Modified date

tomtollafield:
- Add Created Date and Last Modified date to properties collection for all nodes

### DIFF
--- a/DFC.ServiceTaxonomy.GraphSync/GraphSyncers/BagPartGraphSyncer.cs
+++ b/DFC.ServiceTaxonomy.GraphSync/GraphSyncers/BagPartGraphSyncer.cs
@@ -43,8 +43,8 @@ namespace DFC.ServiceTaxonomy.GraphSync.GraphSyncers
 
                 string contentType = contentItem!["ContentType"]!.ToString();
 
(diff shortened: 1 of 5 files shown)


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/BagPartGraphSyncer.cs
```diff
@@ -42,8 +42,12 @@ public BagPartGraphSyncer(IContentDefinitionManager contentDefinitionManager, IS
                 var graphSyncer = _serviceProvider.GetRequiredService<IGraphSyncer>();
 
                 string contentType = contentItem!["ContentType"]!.ToString();
+
+                DateTime? createdDate = DateTime.Parse(contentItem["CreatedDate"]!.ToString());
```
lazcool: As `DateTime.Parse` throws if there's a problem with the input string, we should be able to place the result in a non-nullable DateTime and remove the checks for null later on.

tomtollafield: Updated

PR #50, opened 2020-03-09
Initial graph sync audit workflow

sturatcliffe:


DFC.ServiceTaxonomy.GraphSync/Activities/AuditSyncIssuesTask.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            _session = session;
+            _graphDatabase = graphDatabase;
+            _upsertGraphSyncer = upsertGraphSyncer;
+            T = localizer;
+            _partSyncers = partSyncers.ToDictionary(x => x.PartName ?? "Eponymous");
+            _syncFailedContentItems = new List<ContentItem>();
+            _contentTypes = contentDefinitionManager
+                .ListTypeDefinitions()
+                .Where(x => x.Parts.Any(p => p.Name == "GraphSyncPart"))
```
lazcool: We should use `nameof` to make the code refactor friendly.

DFC.ServiceTaxonomy.GraphSync/Activities/AuditSyncIssuesTask.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        private readonly ISession _session;
+        private readonly IGraphDatabase _graphDatabase;
+        private readonly Dictionary<string, IContentPartGraphSyncer> _partSyncers;
+        private readonly IUpsertGraphSyncer _upsertGraphSyncer;
+        private IStringLocalizer T { get; }
+
+        public override string Name => nameof(AuditSyncIssuesTask);
+        public override LocalizedString DisplayText => T["Identify sync issues and retry"];
+        public override LocalizedString Category => T["National Careers Service"];
```
lazcool: The category for our (generic) graph sync components has been changed to "Graph", we should have this match.

DFC.ServiceTaxonomy.GraphLookup/GraphSyncers/GraphLookupPartGraphSyncer.cs
```diff
@@ -47,6 +48,32 @@ public class GraphLookupPartGraphSyncer : IContentPartGraphSyncer
...
         }
 
+        public Task<bool> VerifySyncComponent(ContentItem contentItem, INode node, ContentTypePartDefinition contentTypePartDefinition, IEnumerable<IRelationship> relationships, IEnumerable<INode> destNodes)
+        {
+            var nodes = (JArray)contentItem.Content.GraphLookupPart.Nodes;
+            var relationshipType = (string)contentTypePartDefinition.Settings["GraphLookupPartSettings"]!["RelationshipType"]!;
+
+            foreach (var n in nodes)
+            {
+                var destNode = destNodes.SingleOrDefault(x => (string)x.Properties["uri"] == (string)n["Id"]!);
```
lazcool: We should inject `IGraphSyncPartIdProperty` and use the `Name` property from it, instead of hard-coding "uri".
Eventually, the name of the id property should come from a settings value, but we should use the interface for now.

DFC.ServiceTaxonomy.GraphSync/Queries/MatchNodeWithAllOutgoingRelationshipsQuery.cs
```diff
@@ -0,0 +1,35 @@
...
+        public void CheckIsValid()
+        {
+            // nothing to check, all properties are non-nullable
+        }
+
+        public Query Query
+        {
+            get
+            {
+                return new Query($"match (s:ncs__{ContentType} {{ uri: '{Uri}' }})-[r]->(d) return s, r, d");
```
lazcool: Does this need to match on the node and have an optional match on the relationship (in a similar fashion to the  query in `DeleteNodeCommand`)?

lazcool:
Shouldn't there be an update to the recipe to add a new workflow?

DFC.ServiceTaxonomy.GraphSync/Activities/AuditSyncIssuesTask.cs
```diff
@@ -0,0 +1,144 @@
...
+                .Query<ContentItem, ContentItemIndex>(x =>
+                    x.ContentType.IsIn(_contentTypes.Keys) && x.Latest && x.Published && (x.CreatedUtc >= log.LastSynced ||  x.ModifiedUtc >= log.LastSynced))
+                .ListAsync();
+
+            foreach (var contentItem in contentItems)
+            {
+                await CheckIfContentItemSynced(contentItem);
+            }
+
+            foreach (var contentItem in _syncFailedContentItems)
```
lazcool: If the content item can't sync correctly for some non-transient reason and `SyncToGraph` doesn't throw, this looks like it could loop forever.
We should probably have a max try count for syncing a single content item.

DFC.ServiceTaxonomy.GraphSync/Activities/AuditSyncIssuesTask.cs
```diff
@@ -0,0 +1,144 @@
...
+                    await _mergeGraphSyncer.SyncToGraph(contentItem.ContentType, contentItem.Content,
+                        contentItem.CreatedUtc, contentItem.ModifiedUtc);
+
+                    _syncFailedContentItems.Remove(contentItem);
+
+                    await CheckIfContentItemSynced(contentItem);
+                }
+                catch
+                {
+                    //I don't want to do anything here, why won't this build without this comment?
```
lazcool: We should log the details of any exception.
We should also probably log all the content items that require a resync.

DFC.ServiceTaxonomy.GraphLookup/GraphSyncers/GraphLookupPartGraphSyncer.cs
```diff
@@ -47,6 +49,32 @@ public class GraphLookupPartGraphSyncer : IContentPartGraphSyncer
             return emptyResult;
         }
 
+        public Task<bool> VerifySyncComponent(ContentItem contentItem, INode node, ContentTypePartDefinition contentTypePartDefinition, IEnumerable<IRelationship> relationships, IEnumerable<INode> destNodes)
+        {
+            var nodes = (JArray)contentItem.Content.GraphLookupPart.Nodes;
+            var relationshipType = (string)contentTypePartDefinition.Settings["GraphLookupPartSettings"]!["RelationshipType"]!;
+
+            foreach (var n in nodes)
```
lazcool: Only a matter of style, but we shouldn't have 1 letter variable names (outside of lambdas).

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Interfaces/IContentPartGraphSyncer.cs
```diff
@@ -14,5 +16,12 @@ public interface IContentPartGraphSyncer
             IMergeNodeCommand mergeNodeCommand,
             IReplaceRelationshipsCommand replaceRelationshipsCommand,
             ContentTypePartDefinition contentTypePartDefinition);
+
+        Task<bool> VerifySyncComponent(
+            ContentItem contentItem,
```
lazcool: What do you think of reordering the parameters to `content, conterntTypePartDefinition, node, relationships, destNodes`, so that the content and graph elements are grouped together?

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/GraphSyncPartGraphSyncer.cs
```diff
@@ -29,5 +32,12 @@ public GraphSyncPartGraphSyncer(IGraphSyncPartIdProperty graphSyncPartIdProperty
 
             return Task.FromResult(Enumerable.Empty<ICommand>());
         }
+
+        public Task<bool> VerifySyncComponent(ContentItem contentItem, INode node, ContentTypePartDefinition contentTypePartDefinition,
+            IEnumerable<IRelationship> relationships, IEnumerable<INode> destNodes)
+        {
+            var uri = node.Properties["uri"];
```
lazcool: This should use the `_graphSyncPartIdProperty` to get the id name (and can be used to extract the value from the graph sync part), similarly to its use in `AddSyncComponents` above.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
```diff
@@ -114,6 +120,68 @@ public class EponymousPartGraphSyncer : IContentPartGraphSyncer
...
+                    if (contentCount != relationshipCount)
+                    {
+                        return false;
+                    }
+
+                    foreach (var item in contentItemIds)
+                    {
+                        var contentItemId = (string)item!;
+
+                        var destContentItem = await _session
```
lazcool: It's probably more appropriate to use `_contentManager.GetAsync` here instead.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
```diff
@@ -114,6 +120,68 @@ public class EponymousPartGraphSyncer : IContentPartGraphSyncer
...
+
+                        if (relationship == null)
+                        {
+                            return false;
+                        }
+                    }
+                }
+                else
+                {
+                    var contentItemValue = value["Text"] ?? value["Html"] ?? value["Value"];
```
lazcool: We should handle link fields in this section too.

lazcool: We'll have to be careful we don't fail validation for a field we don't sync.
We should also check `AddSyncComponents` works correctly for fields we haven't explicitly added support for.
Probably outside the scope of this PR though, so perhaps we just add a todo for now.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
```diff
@@ -114,6 +120,68 @@ public class EponymousPartGraphSyncer : IContentPartGraphSyncer
...
+                    var nodePropertyValue = node.Properties[$"ncs__{field.Name}"];
+
+                    if (Convert.ToString(contentItemValue) != Convert.ToString(nodePropertyValue))
```
lazcool: It'd probably be safer to handle numeric fields explicitly, rather than relying on a string comparison.
From memory, numeric fields with scale 0 are represented as floats in the content field, so show as e.g. "1.0", but we store them as integers in the graph, so would have the string "1".
It might actually work as is, but I think it would still be worth considering handling numerics explicitly.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
```diff
@@ -114,6 +120,68 @@ public class EponymousPartGraphSyncer : IContentPartGraphSyncer
...
+                    {
+                        var contentItemId = (string)item!;
+
+                        var destContentItem = await _session
+                            .Query<ContentItem, ContentItemIndex>(x =>
+                                x.ContentItemId == contentItemId).FirstOrDefaultAsync();
+
+                        var destUri = (string)destContentItem.Content.GraphSyncPart.Text;
+
+                        var destNode = destNodes.SingleOrDefault(n => (string)n.Properties["uri"] == destUri);
```
lazcool: We should get the id fields ("uri") from `IGraphSyncPartIdProperty`.

DFC.ServiceTaxonomy.Editor/Startup.cs
```diff
@@ -11,15 +10,14 @@ public void ConfigureServices(IServiceCollection services)
...
-        
+
         public void Configure(IApplicationBuilder app, IHostEnvironment env)
         {
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();
             }
 
-            app.UseStaticFiles();
```
lazcool: We'll probably need to serve static files for the visualisation module, although the PR for that could add it back if it needs it.

DFC.ServiceTaxonomy.GraphSync/Activities/AuditSyncIssuesTask.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                var contentDefinition = _contentTypes[contentItem.ContentType];
+
+                var sourceNode = results.Select(x => x[0]).Cast<INode>().First();
+                var relationships = results.Select(x => x[1]).Cast<IRelationship>().ToList();
+                var destNodes = results.Select(x => x[2]).Cast<INode>().ToList();
+
+                foreach (var part in contentDefinition.Parts)
+                {
+                    if (!_partSyncers.TryGetValue(part.PartDefinition.Name, out var partSyncer))
```
lazcool: We should only use the eponymous graph syncer if it's definitely the eponymous part (i.e. either set the _partSyncers key to the name of the content type, or try getting "Eponymous" if the part name is the same as the content type name).
By using the eponymous part syncer for all parts without a syncer, it'll have undefined behaviour if the content type contains a part there isn't a graph syncer for.

DFC.ServiceTaxonomy.GraphSync/Activities/AuditSyncIssuesTask.cs
sturatcliffe: Can't see why it would loop forever? It should only try once per content item in the sync failed list, and anything left in there at the end is still failing?

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
sturatcliffe: Not sure what you mean - what fields don't we sync?

DFC.ServiceTaxonomy.Editor/Startup.cs
sturatcliffe: I removed as I vaguely remembered adding it for something I was playing around with, don't think it's needed right now!

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
lazcool: Currently we don't sync fields such as Boolean and DateTime.
We need to ensure the validation doesn't fail if a content type contains any fields we don't (yet) sync, as it's valid for them to be in the ContentItem but not the node properties.

DFC.ServiceTaxonomy.GraphSync/Activities/AuditSyncIssuesTask.cs
lazcool: What I was thinking was this scenario...
```
foreach (var contentItem in _syncFailedContentItems)
{
  ...
  _syncFailedContentItems.Remove(contentItem);
  ...
  await CheckIfContentItemSynced(contentItem); => _syncFailedContentItems.Add(contentItem);
}
```
Googling as to whether that would result in an infinite loop, it looks like mutating, especially removing items from a collection that is being iterated over with a foreach, might be a bad idea anyway, although it looks like the behavior might have changed at some point.
Would you mind checking..
1) if mutating the collection within the foreach is ok?
2) that it won't loop forever if you remove then add back an item into the _syncFailedContentItems collection

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/BagPartGraphSyncer.cs
```diff
@@ -81,10 +82,31 @@ public BagPartGraphSyncer(IContentDefinitionManager contentDefinitionManager, IS
...
+            IEnumerable<IRelationship> relationships,
+            IEnumerable<INode> destNodes)
+        {
+            var graphSyncValidator = _serviceProvider.GetRequiredService<IGraphSyncValidator>();
+
+            var contentItems = contentItem.Content[contentTypePartDefinition.Name]["ContentItems"].ToObject<IEnumerable<ContentItem>>();
+
+            foreach (var bagPartContentItem in contentItems)
+            {
+                if (await graphSyncValidator.CheckIfContentItemSynced(bagPartContentItem))
```
lazcool: Apologies if you're still working on this 😄 , but the foreach will never get part the first iteration.

PR #62, opened 2020-03-17
DST-189 Log to AppInsights

sturatcliffe:


DFC.ServiceTaxonomy.Editor/Startup.cs
```diff
@@ -1,13 +1,30 @@
 using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 
 namespace DFC.ServiceTaxonomy.Editor
 {
     public class Startup
     {
+        public Startup(IHostEnvironment env)
```
lazcool: I thought OC already did this for us?

sturatcliffe: it might do, but how do we access the Configuration instance to access the env variables?

PR #73, opened 2020-03-20
DST-235 Resync deleted content items

sturatcliffe:


lazcool:
Implemented in a newer PR.

PR #77, opened 2020-03-24
Allow ints in add numeric properties

tomtollafield:


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
```diff
@@ -201,8 +203,8 @@ private async Task AddNumericProperties(IMergeNodeCommand mergeNodeCommand, stri
             var fieldDefinition = contentTypePartDefinition.PartDefinition.Fields.First(f => f.Name == fieldName);
             var fieldSettings = fieldDefinition.GetSettings<NumericFieldSettings>();
 
-            string propertyName = await _graphSyncHelper!.PropertyName(fieldName);
-            if (fieldSettings.Scale == 0)
+            string propertyName = $"{NcsPrefix}{fieldName}";
```
lazcool: We should use GraphSyncHelper now and not hard code ncs prefix.

PR #78, opened 2020-03-24
Feature/dst 99 remove fields event

tomtollafield:


DFC.ServiceTaxonomy.GraphSync/Activities/RemoveFieldFromContentItemsTask.cs
```diff
@@ -0,0 +1,82 @@
...
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+        public override async Task<ActivityExecutionResult> ExecuteAsync(WorkflowExecutionContext workflowContext, ActivityContext activityContext)
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+        {
+            string? contentTypeToSync, fieldToRemove;
+            ValidateParameters(workflowContext, out contentTypeToSync, out fieldToRemove);
+
+            _contentDefinitionService.RemoveFieldFromPart(fieldToRemove, contentTypeToSync);
+
+            var b = _contentManager.GetPartDefinition(contentTypeToSync);
```
lazcool: Would you mind explaining this check for me?
Also, could we have a more descriptive variable name and is `InvalidCastException` an appropriate exception to throw?

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
```diff
@@ -78,9 +78,20 @@ public EponymousPartGraphSyncer(IContentManager contentManager)
...
+
+                string? keyName = ((JProperty)field).Name;
+
                 if (firstProperty == null)
                     continue;
 
+                if (!contentTypePartDefinition.PartDefinition.Fields.Any(z => z.Name.Equals(keyName, StringComparison.OrdinalIgnoreCase)))
+                {
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+                    mergeNodeCommand.Properties.Add(await _graphSyncHelper!.PropertyName(keyName), null);
```
lazcool: Is this really necessary?
I think properties with a null value aren't actually created on the node in neo.
If there is a good reason, we'll have to make sure the change doesn't go missing during merge, as the epoonymous graph syncer has been refactored in an outstanding pr.

DFC.ServiceTaxonomy.GraphSync/Activities/RemoveFieldFromContentItemsTask.cs
tomtollafield: Aha! This is POC code that needs removing, I was trying to see which point the cache wasn't being updated and missed removing this

DFC.ServiceTaxonomy.GraphSync/Managers/CustomContentDefinitionManager.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// </summary>
+        private ContentDefinitionRecord LoadContentDefinitionRecord() =>
+            _contentDefinitionStore.LoadContentDefinitionAsync().GetAwaiter().GetResult();
+
+        private ContentTypeDefinition Build(ContentTypeDefinitionRecord source, IList<ContentPartDefinitionRecord> partDefinitionRecords)
+        {
+            if (source == null)
+            {
+#pragma warning disable CS8603 // Possible null reference return.
+                return null;
```
lazcool: I think the warning is valid here. The method returns a non-nullable reference type, but the method returns null.
As the method is private, we should be able to return `ContentTypeDefinition?`.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/EponymousPartGraphSyncer.cs
tomtollafield: Now that the cache is clearing we may be able to simply continue here - I shall give it a go. The problem I had was that not passing the property as null to Neo, it was still appearing, but that could be due to caching.

DFC.ServiceTaxonomy.GraphSync/Services/CustomContentDefinitionService.cs
```diff
@@ -131,7 +131,16 @@ public IEnumerable<EditTypeViewModel> LoadTypes()
 
         public void RemoveFieldFromPart(string fieldName, string partName)
         {
-            _orchardCoreContentDefinitionService.RemoveFieldFromPart(fieldName, partName);
+            var typeBeingUpdated = _contentDefinitionManager.GetTypeDefinition(partName);
+
+            if (typeBeingUpdated != null && typeBeingUpdated.Parts.Any(x => x.Name == "GraphSyncPart"))
```
lazcool: Could you change this to `nameof(GraphSyncPart)` to make it more refactor friendly.

PR #79, opened 2020-03-24
NCSD-3125: Adds OwlDataGenerator app settings

thelious1974:


Resources/AzureDevOps/VariableTemplates/SharedEnvironmentVariables.yml
```diff
@@ -0,0 +1,15 @@
+﻿variables:
```
thelious1974: The original case said these values *may* change across environments, so I've placed these in a variable set shared between all environments. 

If we need to change a specific value for a specific environment, we can move it to the environment specific sets, but for now this will suffice

PR #89, opened 2020-03-30
Ncsd 3193 add custom hostname

NickGraham101:


Resources/ArmTemplates/template.json
```diff
@@ -172,6 +191,33 @@
                 "[variables('AppServiceAppInsightsName')]"
             ]
         },
+        {
+            "name": "editor-certificate",
```
thelious1974: q: Does the app service deployment need to be dependant on this?

NickGraham101: Yes, added dependency

PR #117, opened 2020-03-31
Include a custom script on every page to reorder the admin menu items…

sturatcliffe:
… until an OC issue is resolved.


DFC.ServiceTaxonomy.Editor/wwwroot/ncs.js
```diff
@@ -0,0 +1,10 @@
+﻿$(function () {
+    var ul = $("ul#adminMenu li#new ul");
```
lazcool, line 2: Could this be simplified to `#new ul`?

DFC.ServiceTaxonomy.Editor/wwwroot/ncs.js
```diff
@@ -0,0 +1,10 @@
...
+    var li = ul.children("li");
+
+    li.detach().sort(function (a, b) {
+        return (a.innerText > b.innerText) - (a.innerText < b.innerText);
```
lazcool, line 6: This looks a bit odd to me, with the boolean arithmetic.
Perhaps an easier to follow bit of code would be better, even it it takes a few more lines?

DFC.ServiceTaxonomy.Editor/wwwroot/ncs.js
sturatcliffe: It could, but I like to be specific in case there are other elements with an ID of "new" - doesn't seem overly unique! What do you think?

lazcool: Id's must be unique within a html document.
The selector is not performance critical, so it doesn't matter too much, its just a little bit tidier.

sturatcliffe: Done!

DFC.ServiceTaxonomy.Editor/wwwroot/ncs.js
sturatcliffe: Done!

PR #133, opened 2020-04-02
Feature/dst 93 multi instance setup

tomtollafield:


lazcool:
We could do with a way to run a query against all the graphs (or a graph at a time) in `NeoGraphDatabase`, for the sync validation to use.
Also we could distribute the queries across the instances to distribute load (we could have a simple round-robin).

DFC.ServiceTaxonomy.Neo4j/Extensions/DriverExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+            {
+                throw new InvalidOperationException("More than 1 enabled Primary Driver defined in Neo4JConfiguration");
+            }
+
+            if(!primaryDrivers.Any())
+            {
+                throw new InvalidOperationException("No enabled Primary Driver defined in Neo4JConfiguration");
+            }
+
+            return primaryDrivers.FirstOrDefault().Driver;
```
lazcool, line 25: Don't think we need `FirstOrDefault()` here, the above code already ensure's there's 1, so `First()` should be fine. (Also means we don't have to worry about the null return.)

DFC.ServiceTaxonomy.Neo4j/Extensions/DriverExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        public static IEnumerable<INeoDriver> AllDrivers(this IEnumerable<INeoDriver> drivers)
```
lazcool, line 28: Does this add much value?

DFC.ServiceTaxonomy.Neo4j/Models/NeoDriver.cs
```diff
@@ -0,0 +1,19 @@
...
+        public NeoDriver(string type, IDriver driver, string? uri)
+        {
+            this.Driver = driver;
+            this.Type = type;
+            this.Uri = uri;
+        }
+
+        public IDriver Driver { get; set; }
+        public string Type { get; set; }
+        public string? Uri { get; set; }
```
lazcool, line 17: A null Uri doesn't seem to have any significance, so can we change the type here (and in the ctor params & interface) to a non-nullable string?

DFC.ServiceTaxonomy.Neo4j/Services/NeoDriverBuilder.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            _neo4JConfigurationOptions = neo4jConfigurationOptions;
+            _logger = logger;
+        }
+        public IEnumerable<INeoDriver> Build()
+        {
+            //TODO: GroupBy clause shouldn't be needed, but configuration provides duplicates for some reason
+            //A driver defined as primary is used for queries/reads
+            //Wrap the driver creation process to enable mocking
+            return _neo4JConfigurationOptions.CurrentValue.Endpoints.Where(x => x.Enabled).GroupBy(y => y.Uri).Select(z => new NeoDriver(z.FirstOrDefault().Primary ? "Primary" : "Secondary", GraphDatabase.Driver(
```
lazcool, line 26: Could we check if `z.FirstOrDefault()` returns null and throw a meaningful exception? (It'd be nice if we threw helpful exceptions for missing config in general.)

DFC.ServiceTaxonomy.Neo4j/Services/NeoGraphDatabase.cs
```diff
@@ -29,17 +30,17 @@ public class NeoGraphDatabase : IGraphDatabase, IDisposable
...
             //  driver is thread-safe, while the session or the transaction is not thread-safe.
             //todo: add configuration/settings menu item/page so user can enter this
-            var neo4jConfiguration = neo4jConfigurationOptions.CurrentValue;
-
-            _driver = GraphDatabase.Driver(
-                neo4jConfiguration.Endpoint.Uri,
-                AuthTokens.Basic(neo4jConfiguration.Endpoint.Username, neo4jConfiguration.Endpoint.Password),
-                o => o.WithLogger(neoLogger));
+            _drivers = driverBuilder.Build();
+            _currentDriver = 0;
```
lazcool, line 34: Don't think this is necessary.

DFC.ServiceTaxonomy.Neo4j/Services/NeoGraphDatabase.cs
```diff
@@ -29,17 +30,17 @@ public class NeoGraphDatabase : IGraphDatabase, IDisposable
...
         }
 
         public async Task<List<T>> Run<T>(IQuery<T> query)
         {
-            IAsyncSession session = _driver.AsyncSession();
+            var neoDriver = GetNextDriver();
+
+            _logger.LogInformation($"Executing Read commands to: {neoDriver.Uri}");
```
lazcool, line 41: Could this log 'query' rather than 'write commands'?

DFC.ServiceTaxonomy.Neo4j/Services/NeoGraphDatabase.cs
```diff
@@ -56,12 +57,20 @@ public async Task<List<T>> Run<T>(IQuery<T> query)
...
+
+        private async Task ExecuteTransaction(ICommand[] commands, INeoDriver neoDriver)
+        {
+            IAsyncSession session = neoDriver.Driver.AsyncSession();
             try
             {
                 // transaction functions auto-retry
                 //todo: configure retry? timeout? etc.
 
+                _logger.LogInformation($"Executing Write commands to: {neoDriver.Uri}");
```
lazcool, line 72: Could this log just 'commands' rather than 'Write commands'? (Extra picky i know 😄!)

DFC.ServiceTaxonomy.UnitTests/Neo4j/Extensions/DriverExtensionsTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var listOfDrivers = new List<NeoDriver>() {
+                new NeoDriver("Primary", A.Fake<IDriver>(), "http://my-primary-driver"),
+                new NeoDriver("Secondary", A.Fake<IDriver>(), "http://my-secondary-driver"),
+                new NeoDriver("Secondary", A.Fake<IDriver>(), "http://my-secondary-2-driver"),
+            };
+
+            var primaryDriver = listOfDrivers.PrimaryDriver();
+
+            Assert.NotNull(primaryDriver);
```
lazcool, line 26: Could this check that the returned driver _is_ the primary driver (in addition to testing that _a_ driver is returned)?

DFC.ServiceTaxonomy.UnitTests/Neo4j/Extensions/DriverExtensionsTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        [Fact]
+        public void DriverExtensions_WhenPassedCollectionWithoutPrimaryDriver_ThrowsException()
+        {
+            var listOfDrivers = new List<NeoDriver>() {
+                new NeoDriver("Secondary", A.Fake<IDriver>(), "http://my-secondary-driver"),
+                new NeoDriver("Secondary", A.Fake<IDriver>(), "http://my-secondary-2-driver"),
+            };
+
+            var exception = Assert.Throws<InvalidOperationException>(listOfDrivers.PrimaryDriver);
```
lazcool, line 37: There's a draft pr that enables sonar analysis on the unit test project, which will complain about `exception` not being used.

DFC.ServiceTaxonomy.UnitTests/Neo4j/Services/NeoDriverBuilderTests.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public class NeoDriverBuilderTests
+    {
+        private readonly IOptionsMonitor<Neo4jConfiguration> _optionsMonitor;
+
+        public NeoDriverBuilderTests()
+        {
+            _optionsMonitor = A.Fake<IOptionsMonitor<Neo4jConfiguration>>();
+            var neo4jConfiguration = A.Fake<Neo4jConfiguration>();
+            neo4jConfiguration.Endpoints = new System.Collections.Generic.List<EndpointConfiguration> { new EndpointConfiguration { Uri = "bolt://example.com", Enabled = true, Primary = true }, new EndpointConfiguration { Uri = "bolt://anotherexample.com", Enabled = true, Primary = false } };
```
lazcool, line 21: Is there a `List` in another namespace? If not, could we remove the namespace prefix?
Also, could we split the line up a bit?

DFC.ServiceTaxonomy.Neo4j/Configuration/Neo4jConfiguration.cs
```diff
@@ -1,9 +1,11 @@
+using System.Collections.Generic;
+
 namespace DFC.ServiceTaxonomy.Neo4j.Configuration
 {
 #pragma warning disable S101 // Types should be named in PascalCase
     public class Neo4jConfiguration
 #pragma warning restore S101 // Types should be named in PascalCase
     {
-        public EndpointConfiguration Endpoint { get; } = new EndpointConfiguration();
+        public List<EndpointConfiguration> Endpoints { get; set; } = new List<EndpointConfiguration>();
```
lazcool, line 9: Would you mind updating the `readme.md` with the new config?

DFC.ServiceTaxonomy.Neo4j/Extensions/DriverExtensions.cs
tomtollafield: Driver extensions is no longer relevant since we added round-robin and is being removed

DFC.ServiceTaxonomy.Neo4j/Extensions/DriverExtensions.cs
tomtollafield: Driver extensions is no longer relevant since we added round-robin and is being removed

DFC.ServiceTaxonomy.Neo4j/Models/NeoDriver.cs
tomtollafield: Endpoint configuration that is passed in is a nullable string type so followed convention

DFC.ServiceTaxonomy.Neo4j/Services/NeoGraphDatabase.cs
tomtollafield: The class is a singleton so current driver is set to 0, and is set back to 0 in the next driver method

DFC.ServiceTaxonomy.UnitTests/Neo4j/Extensions/DriverExtensionsTests.cs
tomtollafield: Driver Extensions being removed, but yeah valid point :)

DFC.ServiceTaxonomy.Neo4j/Models/NeoDriver.cs
lazcool: I know, but if we can remove the nullable here, only `NeoDriverBuilder` has to handle null before it builds `NeoDriver`, whereas if it stays nullable here, all the consumers of `NeoDriver` will have to handle a null `Uri`.

DFC.ServiceTaxonomy.Neo4j/Services/NeoGraphDatabase.cs
lazcool: I believe fields uninitialised in the constructor are set to their default values (i.e. 0).

DFC.ServiceTaxonomy.Neo4j/DFC.ServiceTaxonomy.Neo4j.csproj
```diff
@@ -17,4 +17,8 @@
       </PackageReference>
     </ItemGroup>
 
+    <ItemGroup>
+      <Folder Include="Extensions\" />
```
lazcool, line 21: Don't think this is needed.

PR #148, opened 2020-04-08
DST-302 Sync Validation

lazcool:


DFC.ServiceTaxonomy.GraphSync/Permissions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using OrchardCore.Security.Permissions;
+
+namespace DFC.ServiceTaxonomy.GraphSync
+{
+    public class Permissions : IPermissionProvider
+    {
+        public static readonly Permission AdministerGraphs = new Permission("AdministerGraphs", "Administer graphs");
```
tomtollafield, line 9: Could do with extracting the hard-coded values in permissions into a central class so we can use and reference them elsewhere to avoid typos and promote reuse

DFC.ServiceTaxonomy.GraphSync/Controllers/GraphSyncController.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning(e, "User triggered sync validation failed: {e}");
+                _notifier.Add(NotifyType.Error, new LocalizedHtmlString(nameof(GraphSyncController), $"Unable to verify graph sync."));
+            }
+            return View(new TriggerSyncValidationViewModel
+            {
+                ValidateAndRepairResult = validateAndRepairResult
+            });
```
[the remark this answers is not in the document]
lazcool: Yes, the view explicitly checks that `Model.ValidateAndRepairResult` is not null.

DFC.ServiceTaxonomy.GraphSync/Permissions.cs
lazcool: If you look at an [example of checking the permission](https://github.com/SkillsFundingAgency/dfc-servicetaxonomy-editor/pull/148/files#diff-81a28a57b72e94971dd37ea003e4e3e6), notice that the class itself is used, rather than the strings.

The strings are only for display in the Security admin section.

PR #152, opened 2020-04-15
Features/synonyms from cypher

tomtollafield:


DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
```diff
@@ -1,9 +1,50 @@
...
+            stream.Position = 0;
+            return new FileStreamResult(stream, "text/plain")
+            {
+                FileDownloadName = filename
+            };
+        }
+
+        public static Stream GenerateStreamFromString(string s)
+        {
+            var stream = new MemoryStream();
```
lazcool, line 40: `MemoryStream` and `StreamWriter` should be disposed.

DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
```diff
@@ -1,9 +1,50 @@
-﻿using Microsoft.AspNetCore.Mvc;
+﻿using System;
+using System.IO;
+using System.Text;
+using DFC.ServiceTaxonomy.GraphSync.Services.Interface;
+using Microsoft.AspNetCore.Mvc;
 
 namespace DFC.ServiceTaxonomy.GraphSync.Controllers
 {
     public class HomeController : Controller
```
lazcool, line 9: I deleted the `HomeController` in my PR and created a `GraphSyncController` 😄 - not that it really matters.

DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
tomtollafield: Thanks - noticed there is some legacy code here that needs removing too

DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
```diff
@@ -1,9 +1,50 @@
...
     {
+        private readonly ISynonymService _synonymService;
+
+        public HomeController(ISynonymService synonymService)
+        {
+            _synonymService = synonymService ?? throw new ArgumentNullException(nameof(synonymService));
```
lazcool, line 15: The null check is not really necessary, as the controller won't be created unless `ISynonymService` is available (unless the controller is being used outside of asp.net I suppose). Not that the check will do any harm.

DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
```diff
@@ -1,9 +1,50 @@
...
+            var stream = new MemoryStream(buffer: Encoding.UTF8.GetBytes(sb.ToString()));
+
+            stream.Position = 0;
+            return new FileStreamResult(stream, "text/plain")
+            {
+                FileDownloadName = filename
+            };
+        }
+
+        public static Stream GenerateStreamFromString(string s)
```
lazcool, line 38: This method doesn't seen to be used?

tomtollafield: Yeah I replied above that I'm just removing it - legacy code :) thanks for the spot

DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
tomtollafield: As this is DId, if there is no service collection entry this should throw? Or is it because it's the only parameter in the constructor?

DFC.ServiceTaxonomy.GraphSync/Services/SynonymService.cs
```diff
@@ -0,0 +1,42 @@
...
+        private readonly IGraphDatabase _neoGraphDatabase;
+
+        public SynonymService(IGraphDatabase neoGraphDatabase)
+        {
+            _neoGraphDatabase = neoGraphDatabase ?? throw new ArgumentNullException(nameof(neoGraphDatabase));
+        }
+
+        public IEnumerable<string> GetSynonyms(string node)
+        {
+            var query = new MatchSynonymsQuery();
```
lazcool, line 21: I don't think we need to create an empty `MatchSynonymsQuery`, which gets thrown away if node is skill or occupation.

Then we could remove the constructor of `MatchSynonymsQuery` with no parameters and there would be no need to call `this.CheckIsValid()` in the query (which it should do, if there's a possibility that the params haven't been supplied).

If node is not skill or occupation we should properly handle it, rather than execute a query with empty string params.

DFC.ServiceTaxonomy.GraphSync/Queries/MatchSynonymsQuery.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            return new List<string>();
+            //:ncs__hasAltLabel|:ncs__hasPrefLabel|:ncs__hasHiddenLabel
+        }
+
+        public Query Query
+        {
+            get
+            {
+                return new Query($"MATCH(o:{FirstNodeLabel})-[{RelationshipLabels}]- (l:{SecondNodeLabel}) WITH {{ label: o.{PropertyValue}, values: REDUCE(result = o.{PropertyValue}, s in collect(l.{PropertyValue}) | result + \",\" + s) }} AS synonyms WITH {{ Results: collect(synonyms.values)}} as results RETURN results");
```
lazcool, line 49: We should add a call to `this.CheckIsValid();` first, as the relationship labels might not be supplied to the constructor (assuming the constructor with no params is removed), and we should check for that condition in `ValidationErrors`.

DFC.ServiceTaxonomy.GraphSync/Queries/MatchSynonymsQuery.cs
```diff
@@ -0,0 +1,59 @@
...
+            FirstNodeLabel = firstNodeLabel;
+            SecondNodeLabel = secondNodeLabel;
+            PropertyValue = propertyValue;
+            RelationshipLabels = BuildRelationshipLabels(relationshipLabels);
+        }
+
+        private string BuildRelationshipLabels(string[] relationshipLabels)
+        {
+            for (var i = 0; i < relationshipLabels.Length; i++)
+                relationshipLabels[i] = ":" + relationshipLabels[i];
```
lazcool, line 34: This will leave a bunch of strings that'll need to be garbage collected.
Instead we could use `string.Join("|:", ...` and add an initial `:` before `{RelationshipLabels}` in the `Query` constructor.

DFC.ServiceTaxonomy.GraphSync/Services/SynonymService.cs
tomtollafield: Yeah valid point, I was trying to initialise the query and set it before running it.

A simple refactor would remove this need entirely, updating now

DFC.ServiceTaxonomy.GraphSync/Services/SynonymService.cs
```diff
@@ -0,0 +1,42 @@
...
+
+            if(node.Equals("Skill", StringComparison.CurrentCultureIgnoreCase))
+            {
+                query = new MatchSynonymsQuery("esco__Skill", "ncs__SkillLabel", "skos__prefLabel", "ncs__hasAltLabel", "ncs__hasPrefLabel", "ncs__hasHiddenLabel");
```
lazcool: We should get these values, if we can, from the appropriate settings using `GraphSyncHelper`, rather than hard-coding them.

DFC.ServiceTaxonomy.GraphSync/Services/SynonymService.cs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            if (node.Equals("Occupation", StringComparison.CurrentCultureIgnoreCase))
+            {
+                query = new MatchSynonymsQuery("esco__Occupation", "ncs__OccupationLabel", "skos__prefLabel", "ncs__hasAltLabel", "ncs__hasPrefLabel", "ncs__hasHiddenLabel");
+            }
+
+            var result = _neoGraphDatabase.Run(query).GetAwaiter().GetResult();
```
lazcool, line 37: Wouldn't it be better to make this method and the controller action async and use await instead?

DFC.ServiceTaxonomy.GraphSync/Startup.cs
```diff
@@ -126,6 +128,13 @@ public override void Configure(IApplicationBuilder builder, IEndpointRouteBuilde
                 pattern: "Home/Index",
                 defaults: new { controller = "Home", action = "Index" }
             );
+
+            routes.MapAreaControllerRoute(
```
lazcool, line 132: Do we need to both map the route and use attribute routing?

DFC.ServiceTaxonomy.GraphSync/Services/SynonymService.cs
tomtollafield: Yeah good point, this was made non-async for when it was inside a workflow task

DFC.ServiceTaxonomy.Neo4j/Synonyms/src/main/java/analyzer/SynonymHelper.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.lucene.analysis.standard.StandardFilterFactory;
+import org.apache.lucene.analysis.standard.StandardTokenizerFactory;
+import org.apache.lucene.analysis.synonym.SynonymFilterFactory;
+import java.io.*;
+import org.apache.commons.io.FileUtils;
+import java.util.Scanner;
+
+public class SynonymHelper {
+
+    public SynonymHelper() {
```
lazcool, line 13: Don't think you need to supply an empty constructor.

DFC.ServiceTaxonomy.Neo4j/Synonyms/src/main/java/analyzer/SynonymHelper.java
```diff
@@ -0,0 +1,61 @@
...
+    public Analyzer buildAnalyzer(String settingName) {
+			String neo4jPath = System.getenv("NEO4J_CONF") + "/neo4j.conf";
+			String fileUrl = "";
+			
+			File file = new File(neo4jPath);
+
+			try {
+				Scanner scanner = new Scanner(file);
+
+							 int lineNum = 0;
```
lazcool, line 25: The indentation is a bit funky 😄.

DFC.ServiceTaxonomy.Neo4j/Synonyms/src/main/java/analyzer/SynonymHelper.java
tomtollafield: I'll give it a go - it was complaining it couldn't construct

DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
lazcool: If there's no `ISynonymService` registered, the controller will fail to be injected before the constructor is even called, so `synonymService` should never be null.

DFC.ServiceTaxonomy.GraphSync/Controllers/HomeController.cs
```diff
@@ -1,9 +1,48 @@
...
+
+            try
+            {
+                stream.Position = 0;
+                return new FileStreamResult(stream, "text/plain")
+                {
+                    FileDownloadName = filename
+                };
+            }
+            catch
```
lazcool, line 39: Can't you wrap the stream with an using (or use finally)?

tomtollafield: FileStreamResult disposes the stream passed in within it's internal method. If you wrap it you can get an already disposed exception. If you'd like us to handle the dispose I can change it to a FileResult instead.

https://stackoverflow.com/questions/10418594/with-filestreamresult-how-is-the-memorystream-closed

DFC.ServiceTaxonomy.Neo4j/Synonyms/src/main/java/analyzer/SynonymHelper.java
tomtollafield: Lack of CTRL+K+D in notepad++ for Java :(

DFC.ServiceTaxonomy.GraphSync/Startup.cs
tomtollafield: I'll remove the area route, this was in before the attribute routing

DFC.ServiceTaxonomy.GraphSync/Services/SynonymService.cs
tomtollafield: Source nodes now come from Graph Sync Helper. The rest can be updated later on when GraphSyncHelper is extended

PR #160, opened 2020-04-21
Hotfix/job profile fixes

tomtollafield:


DFC.ServiceTaxonomy.Neo4j/Java Plugins/src/main/java/plugins/FormatLinks.java
```diff
@@ -0,0 +1,33 @@
...
+    @Description("Replaces anchor tags with markup UI expects")
+    public String formatLinks(
+            @Name("input") String input) {
+        if (input == null) {
+            return null;
+        }
+		
+		String pattern = "(<a href=\"([^\"]+)\">([^<]+)</a>)";
+      
+		Pattern r = Pattern.compile(pattern);
```
lazcool, line 23: Would be nice if this could be compiled statically.

PR #211, opened 2020-05-19
DST-369 Convert routes in to content pickers

lazcool:
Also consolidates importing from a single job profile spreadsheet, rather than 2 different versions!


GetJobProfiles/ContentRecipes/SharedContent.recipe.json
```diff
@@ -30,7 +30,7 @@
                         }
                     },
                     "GraphSyncPart": {
-                        "Text": "http://nationalcareers.service.gov.uk/sharedcontent/2def38e0-83ae-4bc1-9350-3fb4b00c7857"
+                        "Text": "http://nationalcareers.service.gov.uk/sharedcontent/c0117ac7-115a-4bc1-9350-3fb4b00c7857"
```
tomtollafield, line 33: I think this will need the Content API Prefix replacement code?

GetJobProfiles/Models/Recipe/Fields/Factories/ContentPickerContentItemFactory.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using GetJobProfiles.Models.Recipe.ContentItems.Base;
+
+namespace GetJobProfiles.Models.Recipe.Fields.Factories
+{
+    public class ContentPickerContentItemFactory<T> where T : ContentItem
+    {
+        private readonly EqualityComparer<T> _comparer;
+        public readonly ConcurrentDictionary<T, string> ItemToUsers;
```
tomtollafield, line 12: Is "ItemToUsers" a typo? Might not be I'm just confused

GetJobProfiles/ContentRecipes/SharedContent.recipe.json
lazcool: So it will 😃

GetJobProfiles/Models/Recipe/Fields/Factories/ContentPickerContentItemFactory.cs
lazcool: Hmm yeah, it's a bit confusing. I was thinking of 'users of the entry route'. Let me think of a better name.

PR #216, opened 2020-05-20
Feature/dst 372 remove ncs prefix

tomtollafield:


DFC.ServiceTaxonomy.Editor/Recipes/Service Taxonomy.recipe.json
```diff
@@ -125,8 +125,8 @@
                                     "Position": "2"
                                 },
                                 "GraphSyncPartSettings": {
-                                    "NodeNameTransform": "$\"ncs__{ContentType}\"",
-                                    "PropertyNameTransform": "$\"ncs__{Value}\"",
+                                    "NodeNameTransform": "$\"{ContentType}\"",
```
lazcool, line 128: Perhaps just `ContentType` would work? (Same with `Value`.)

lazcool:
We should update the optimization `PropertyName` in `GraphSyncHelper` too.

PR #229, opened 2020-05-29
DST-432 Make what visualizer displays generic

lazcool:


lazcool:
Obsoleted by a later PR.

PR #231, opened 2020-06-01
- Lower batch sizes

tomtollafield:


DFC.ServiceTaxonomy.GraphSync/Recipes/Executors/CypherToContentStep.cs
```diff
@@ -96,13 +96,14 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
                     .Where(i => i != null)
                     .Select(i => i!);
 
-                foreach (ContentItem preparedContentItem in preparedContentItems)
-                {
-                    await CreateContentItem(preparedContentItem, cypherToContent.SyncBackRequired);
-                }
+                var contentItems = preparedContentItems.Select(x => CacheContentItem(x, cypherToContent.SyncBackRequired));
```
lazcool, line 99: Would you mind using a foreach to iterate here. The select is not really projecting anything and is instead relying on CacheContentItem's side effect.
Also could it be called something _like_ OptionallyDisableSync, to more clearly describe what it's doing?

It might also be worth batching in the step (perhaps using morelinq), in addition to setting the batch size of the recipe itself, but we can add that later.

PR #232, opened 2020-06-02
Readonly graph sync / automate SASS compilation

sturatcliffe:


.gitignore
```diff
@@ -281,3 +281,6 @@ __pycache__/
 
 #generated localization files
 Localization/
+
+/DFC.ServiceTaxonomy.Theme/wwwroot/Styles/trumbowyg_scoped_govuk_frontend.css
```
lazcool, line 285: Would you mind adding these files back into git, as it's useful to see their history of changes.

DFC.ServiceTaxonomy.GraphSync/Views/GraphSyncPart.Edit.cshtml
```diff
@@ -7,3 +7,13 @@
     <input asp-for="Text" class="form-control content-preview-text" readonly>
     <span class="hint">This URI ID is automatically generated</span>
 </fieldset>
+
+@if (Model.Readonly)
+{
+    <script depends-on="jQuery" at="Foot">
+        $(function () {
+            $('input').attr('readonly', 'readonly');
```
lazcool, line 15: This needs to be more targeted. It makes the filter input in the menu read only too.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/ValidateAndRepairGraph.cs
```diff
@@ -69,10 +69,15 @@ public async Task<ValidateAndRepairResult> ValidateGraph()
 
             foreach (ContentTypeDefinition contentTypeDefinition in syncableContentTypeDefinitions)
             {
-                List<ValidationFailure> syncFailures = await ValidateContentItemsOfContentType(contentTypeDefinition, auditSyncLog.LastSynced, result);
-                if (syncFailures.Any())
+                _graphSyncHelper.ContentType = contentTypeDefinition.Name;
+
+                if (!_graphSyncHelper.GraphSyncPartSettings.PreexistingNode)
```
lazcool, line 74: Although we don't want to validate the properties of preexisting nodes, we do want to validate the node's relationships.
The properties of the occupation and skill nodes are read-only, but the user can change the associated labels, which changes the relationships off the nodes.
So when validating, we need to validate (and repair if necessary) the outgoing relationships from read-only nodes.

sturatcliffe: Ok, as discussed I've just removed the preexisting check, as the repair just hooks back into a mergeNodeCommand which already checks if the node is preexisting and only updates the relationships if it is.

PR #236, opened 2020-06-03
- Update ShellSettings

tomtollafield:


DFC.ServiceTaxonomy.GraphSync/Recipes/Executors/CSharpContentStep.cs
```diff
@@ -36,13 +37,14 @@ public class CSharpContentStep : IRecipeStepHandler
             //todo: rename
             ICypherToContentCSharpScriptGlobals cypherToContentCSharpScriptGlobals,
             IConfiguration configuration,
```
lazcool, line 39: Might as well remove configuration from the c'tors.

PR #242, opened 2020-06-08
add (initial) flow part syncer

lazcool:


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/FlowPartSyncer.cs
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        private async Task AddWidgetsSyncComponents(
+            dynamic content,
+            IReplaceRelationshipsCommand replaceRelationshipsCommand,
+            IGraphSyncHelper graphSyncHelper)
+        {
+            int widgetOrdinal = 0;
+            foreach (JObject? contentItem in content[Widgets])
+            {
```
tomtollafield, line 84: Should never happen but you've got a nullable JObject coming into the foreach and then telling the compiler it will never be null with contentItem! might just want to add a null check at the top.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/FlowPartSyncer.cs
```diff
@@ -0,0 +1,257 @@
...
+                    return (false, $"contained item failed validation: {failureReason}");
+
+                // check expected relationship is in graph
+                var flowContentGraphSyncHelper = _serviceProvider.GetRequiredService<IGraphSyncHelper>();
+
+                flowContentGraphSyncHelper.ContentType = flowPartContentItem.ContentType;
+                string expectedRelationshipType = await RelationshipType(flowContentGraphSyncHelper);
+
+                // keep a count of how many relationships of a type we expect to be in the graph
+                expectedRelationshipCounts.TryGetValue(expectedRelationshipType, out int currentCount);
```
tomtollafield, line 222: Can't see the full stack here but is there a possibility that TryGetValue will not find the expectedRelationshipType?

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/ValidateAndRepairGraph.cs
```diff
@@ -195,22 +192,17 @@ public async Task<(bool validated, string failureReason)> ValidateContentItem(Co
...
-                {
-                    partSyncer = _partSyncers["Eponymous"];
-                }
+                IContentPartGraphSyncer partSyncer = _partSyncers.SingleOrDefault(ps =>
+                    ps.CanHandle(contentTypePartDefinition.ContentTypeDefinition.Name,
+                        contentTypePartDefinition.PartDefinition));
 
                 if (partSyncer == null)
                 {
                     // part doesn't have a registered IContentPartGraphSyncer, so we ignore it
```
tomtollafield, line 201: Could log out ContentTypeDefinition.Name here for diagnostics in case we have something setup incorrectly?

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/FlowPartSyncer.cs
lazcool: While there's nothing stopping an IEnumerator's Current method from returning null, in practice, using standard enumerators and foreach, I think it's pretty safe to ignore the possibility of null.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/FlowPartSyncer.cs
lazcool: Yes it won't find the expected relationship type the first time it encounters that type, but then currentCount will be 0, and the next line adds the expected relationship type with a count of 1.
So basically, not finding it is part of the normal flow and is handled.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/ValidateAndRepairGraph.cs
lazcool: Good idea, added.

PR #255, opened 2020-06-16
Lowercase URI

tomtollafield:


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Helpers/GraphSyncHelper.cs
```diff
@@ -163,7 +163,7 @@ public async Task<string> GenerateIdPropertyValue()
                 $"http://data.europa.eu/esco/{_contentType!.ToLowerInvariant()}/{newGuid}",
 
                 "$\"<<ContentApiPrefix>>/{ContentType.ToLowerInvariant()}/{Value}\"" =>
-                $"{_contentApiBaseUrl}/{_contentType!.ToLowerInvariant()}/{newGuid}",
+                $"{_contentApiBaseUrl}/{_contentType!.ToLowerInvariant()}/{newGuid}".ToLowerInvariant(),
```
lazcool, line 166: The config value and what the optimization creates should match, so we need to change the recipe.
(So any ToLower needs to live within the config value).

GetJobProfiles/Models/Recipe/Parts/GraphSyncPart.cs
```diff
@@ -4,7 +4,7 @@ namespace GetJobProfiles.Models.Recipe.Parts
 {
     public class GraphSyncPart
     {
-        public GraphSyncPart(string contentType) => Text = $"[js:configuration('ContentApiPrefix') + '/{contentType.ToLowerInvariant()}/{Guid.NewGuid()}']";
+        public GraphSyncPart(string contentType) => Text = $"[js:configuration('ContentApiPrefix').toLowerCase() + '/{contentType.ToLowerInvariant()}/{Guid.NewGuid()}']";
```
lazcool, line 7: I suspect this config value invokes the c# interpreter during the import, as there won't be an optimization to catch the values.
Can we check if it does, and if so add an optimization so that the interpreter isn't invoked.
If it does get invoked, not only will it slow the import down, but runs the risk of running out of memory.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Helpers/GraphSyncHelper.cs
tomtollafield: This is only called when you create a new Content Item from inside the editor. ToLowerInvariant() is called to make sure the config value is lowered

GetJobProfiles/Models/Recipe/Parts/GraphSyncPart.cs
tomtollafield: This doesn't seem to trigger the GraphSyncHelper optimization as the constructor is only called when the GetJobProfiles utility is ran. The compiled string then gets baked into each recipes GraphSyncPart.Text property. the js:configuration will be evaluated by the shell when the recipe runs

PR #264, opened 2020-06-18
Feature/dfcc 369 email content type

tomtollafield:


GetJobProfiles/ContentRecipes/EmailTemplates.recipe.json
```diff
@@ -0,0 +1,84 @@
...
+                        }
+                    },
+                    "GraphSyncPart": {
+                        "Text": "http://localhost:7071/api/execute/email/2a090c37-01ca-446a-b60f-0de5ad5321cc"
```
lazcool, line 42: Needs to use `<<ContentApiPrefix>>`. (Same with the item below.)

DFC.ServiceTaxonomy.Editor/Recipes/Service Taxonomy.recipe.json
```diff
@@ -3863,6 +3913,58 @@
...
+                                    "Editor": "Trumbowyg",
+                                    "Position": "3"
+                                },
+                                "HtmlFieldSettings": {},
+                                "HtmlFieldTrumbowygEditorSettings": {
+                                    "Options": "{\r\n    autogrow: true,\r\n    btns: [\r\n        [\"viewHTML\"],\r\n        [\"undo\", \"redo\"],\r\n        [\"formatting\"],\r\n        [\"strong\", \"em\", \"del\"],\r\n        [\"foreColor\", \"backColor\"],\r\n        [\"superscript\", \"subscript\"],\r\n        [\"link\"],\r\n        [\"image\"],\r\n        [\"align\"],\r\n        [\"unorderedList\", \"orderedList\"],\r\n        [\"horizontalRule\"],\r\n        [\"removeformat\"],\r\n        [\"fullscreen\"]\r\n    ],\r\n    btnsDef: {\r\n        align: {\r\n            dropdown: [\"justifyLeft\", \"justifyCenter\", \"justifyRight\", \"justifyFull\"],\r\n            ico: \"justifyLeft\"\r\n        },\r\n        image: {\r\n            dropdown: [\"insertImage\", \"base64\", \"noembed\"],\r\n            ico: \"insertImage\"\r\n        }\r\n    }\r\n}"
```
lazcool, line 3962: I initially thought these settings should be the same as our other trumbowyg editors, but I think default is right, as the email html won't have any of the GDS stylesheets.
You might want to allow <div>'s however, which I think get stripped out when you use the default settings. (Needs semantic: {'div':'div'})

PR #271, opened 2020-06-29
DFCC-542 Draft graph

lazcool:


DFC.ServiceTaxonomy.Editor/DFC.ServiceTaxonomy.Editor.csproj
```diff
@@ -33,15 +33,20 @@
 
   <ItemGroup>
     <None Include="..\.editorconfig" Link=".editorconfig" />
+    <Content Remove="appsettings.Development_template.json" />
```
tomtollafield, line 36: Did you mean to leave this in?

DFC.ServiceTaxonomy.Editor/Startup.cs
```diff
@@ -21,6 +23,14 @@ public void ConfigureServices(IServiceCollection services)
                 options.InstrumentationKey = Configuration["APPINSIGHTS_INSTRUMENTATIONKEY"]);
 
             services.AddOrchardCms();
+
+            services.AddEventGridPublishing(Configuration);
+
+            //todo: do this in each library??? if so, make sure it doesn't add services or config twice
+            services.AddGraphCluster(options =>
```
tomtollafield, line 30: We might want to do a string null check on the section to make sure we're not binding to null. If you already check later on though, ignore this comment.

DFC.ServiceTaxonomy.Editor/DFC.ServiceTaxonomy.Editor.csproj
lazcool: Yes, without it the file gets copied to the bin folder.

DFC.ServiceTaxonomy.Editor/Startup.cs
lazcool: The MSDN states that `GetSection()` returns `null` if the section is missing, except it doesn't, it returns a valid (empty) section instead 😄.
So instead, I've added checks for missing config items in `GraphClusterBuilder` and it now throws a `ConfigurationErrorsException` is there are no endpoints or replica sets configured.

PR #274, opened 2020-07-06
Dfcc 502 limit labels in subset

SimonWren:


GetJobProfiles/Program.cs
```diff
@@ -282,6 +291,10 @@ private static async Task CopyRecipeWithTokenisation(string recipePath, string r
                 recipeName = $"{recipeName}{skip}";
                 tokens["recipeName"] = $"{recipeName}_{_executionId}";
             }
+            else
+            {
+                tokens["recipeName"] = $"{recipeName}_{_executionId}";
```
lazcool, line 296: This line could be placed outside of the `if`, as it's the same in both cases.

PR #277, opened 2020-07-13
Sync with hint

tomtollafield:


DFC.ServiceTaxonomy.Editor/Recipes/Service Taxonomy.recipe.json
```diff
@@ -2522,7 +2522,9 @@
                                     "Editor": "TextArea",
                                     "Position": "1"
                                 },
-                                "TextFieldSettings": {},
+                                "TextFieldSettings": {
+                                    "Hint": "##SyncToArray"
```
lazcool, line 2526: Would you mind adding some hint text - should have been there already.

DFC.ServiceTaxonomy.GraphSync/Extensions/MergeNodeExtensions.cs
```diff
@@ -62,6 +62,24 @@ public static class MergeNodeExtensions
...
+            }
+
+            return new List<T>();
+        }
+
+        public static List<T>? AddArrayProperty<T>(
+           this IMergeNodeCommand mergeNodeCommand,
+           string nodePropertyName,
+           JArray content,
+           string contentPropertyName)
```
lazcool, line 79: Doesn't need this param.

DFC.ServiceTaxonomy.GraphSync/Extensions/MergeNodeExtensions.cs
```diff
@@ -62,6 +62,24 @@ public static class MergeNodeExtensions
         {
             List<T>? values;
             JArray? jarray = (JArray?)content[contentPropertyName];
+
+            if (jarray != null)
```
lazcool, line 66: Looks like this has changed the behaviour of this method.
If jarray is null, then it now returns an empty list, whereas before it returned null (while still returning null if JArrayType == JTokenType.Null, and behaving differently to the AddArrayProperty overload below).
We should return null if jarray is null, which would be best achieved by not checking for != null here and just returning the result of AddArrayProperty direct.

PR #291, opened 2020-07-23
DFCC-668 Delete embedded content items from graph

lazcool:
### Changes
* converts sync into a 2 phase sync across the published and preview graph replica sets. First it checks with all the graph syncers to give them a change to block the sync, before it goes ahead and actually syncs. This mechanism enables us to keep the published and preview graph sets consistent.
* embedded content item parts (e.g. taxonomy terms, flows, bags) now check if there are any incoming relationships on embedded content items that would fail the sync (excluding two way relationships), and block the sync if so.
* publish removed content item event grid events (unpublished, draft-discarded & deleted) with relative urls, as the item can't be fetched and the event could apply to published & preview local stores.


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/MergeGraphSyncer.cs
```diff
@@ -59,94 +82,119 @@ public class MergeGraphSyncer : IMergeGraphSyncer
...
                 }
             }
         }
 
-        private async Task SyncComponentsToGraphReplicaSet(IGraphReplicaSet graphReplicaSet, dynamic graphSyncPartContent)
+        //??
+        // public async Task AddIdSyncComponents(ContentItem contentItem)
+        // {
+        //
+        // }
```
tomtollafield, line 177: Could just delete this bit for now

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/Flow/FlowPartGraphSyncer.cs
```diff
@@ -27,8 +28,16 @@ public class FlowPartGraphSyncer : IContentPartGraphSyncer
             _contentFieldsGraphSyncer = contentFieldsGraphSyncer;
         }
 
+        public async Task AllowSync(JArray? contentItems, IGraphMergeContext context, IAllowSyncResult allowSyncResult)
+        {
+            await Task.WhenAll(
+                _flowPartEmbeddedContentItemsGraphSyncer.AllowSync(contentItems, context, allowSyncResult),
+                _contentFieldsGraphSyncer.AllowSync(contentItems, context, allowSyncResult));
```
tomtollafield, line 35: Is it worth only executing this WhenAll if contentItems is not null? Only a very minor point

lazcool: Oops, this wasn't implementing the correct `AllowSync` signature. Fixed.

PR #295, opened 2020-07-27
Filter by page location, and display default tag on default pages in …

sturatcliffe:
…content item list


DFC.ServiceTaxonomy.PageLocation/ViewModels/PageLocationViewModel.cs
```diff
@@ -0,0 +1,17 @@
...
+using OrchardCore.DisplayManagement.Views;
+
+namespace DFC.ServiceTaxonomy.PageLocation.ViewModels
+{
+    public class PageLocationViewModel : ShapeViewModel
+    {
+        public PageLocationViewModel(ContentItem contentItem)
+        {
+            IsDefault = contentItem.ContentType == "Page" &&
+                Convert.ToBoolean(contentItem.Content.PageLocationPart.DefaultPageForLocation.Value);
```
lazcool, line 12: Would be good to use `nameof` for the part, and a public const string in the part for `DefaultPageForLocation`, to make this refactor friendly.

PR #315, opened 2020-08-07
DFCC-625 Delete field & type

lazcool:
Adds published/preview support for removing a field.
Adds published/preview support for removing a type.

Fixes:
* when removing a field, it would fail if the field wasn't in an eponymous part


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/GraphResyncer.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        private async Task Sync(IEnumerable<ContentItem> contentItems, IGraphReplicaSet graphReplicaSet)
+        {
+            foreach (ContentItem contentItem in contentItems)
+            {
+                IMergeGraphSyncer mergeGraphSyncer = _serviceProvider.GetRequiredService<IMergeGraphSyncer>();
+                //todo: inject?
+                IContentManager contentManager = _serviceProvider.GetRequiredService<IContentManager>();
+
+                await mergeGraphSyncer.SyncToGraphReplicaSetIfAllowed(graphReplicaSet, contentItem, contentManager);
```
tomtollafield, line 53: Is it a big refactor to do constructor injection inside MergeGraphSyncer? Assuming from your comment that it might cause a lot of re-work

lazcool: We can't inject `ContentManager` inside `MergeGraphSyncer` because it causes a circular dependency.
Just testing if we can inject it into `GraphResyncer` instead.

PR #316, opened 2020-08-07
Changed UrlName validation to disallow invalid characters etc

sturatcliffe:


DFC.ServiceTaxonomy.PageLocation/Drivers/PageLocationPartDisplayDriver.cs
```diff
@@ -49,15 +50,30 @@ public override async Task<IDisplayResult> UpdateAsync(PageLocationPart model, I
...
+            }
+
+            if (urlNameIsValid && pageLocation.UrlName!.All(c => !char.IsLetterOrDigit(c)))
+            {
+                updater.ModelState.AddModelError(Prefix, nameof(pageLocation.UrlName), $"'UrlName' must contain at least one alphanumeric character.");
+                urlNameIsValid = false;
             }
 
-            if (pageLocation.UrlName?.IndexOfAny(InvalidCharactersForPath) > -1 || pageLocation.UrlName?.IndexOf(' ') > -1)
+            if (urlNameIsValid && pageLocation.UrlName!.All(c => !char.IsLetter(c)))
```
lazcool, line 73: As this is more restrictive than the previous check, can't we just have this check?

DFC.ServiceTaxonomy.PageLocation/Drivers/PageLocationPartDisplayDriver.cs
```diff
@@ -49,15 +50,30 @@ public override async Task<IDisplayResult> UpdateAsync(PageLocationPart model, I
...
-            if (string.IsNullOrWhiteSpace(pageLocation.UrlName))
+            bool urlNameIsValid = true;
+
+            if (urlNameIsValid && string.IsNullOrWhiteSpace(pageLocation.UrlName))
             {
                 updater.ModelState.AddModelError(Prefix, nameof(pageLocation.UrlName), "A value is required for 'UrlName'");
+                urlNameIsValid = false;
+            }
+
+            if (urlNameIsValid && !Regex.IsMatch(pageLocation.UrlName!, UrlNamePattern))
```
lazcool, line 61: Could this be achieved by adding a `RegularExpression` data annotation attribute on the model?
(Genuinely don't know if it's supported!)

sturatcliffe: Not sure, I'm just following the way they do validation within OC, so potentially not!

PR #321, opened 2020-08-12
DFCC-383 Taxonomy events not firing

sturatcliffe:


DFC.ServiceTaxonomy.Taxonomies/Controllers/AdminController.cs
```diff
@@ -0,0 +1,317 @@
...
+
+                if (taxonomyItems == null)
+                {
+                    parentTaxonomyItem["Terms"] = taxonomyItems = new JArray();
+                }
+
+                taxonomyItems.Add(JObject.FromObject(contentItem));
+            }
+
+            await _contentManager.PublishAsync(taxonomy);
```
lazcool, line 140: Could you please update the Stax startup recipe, so that taxonomies/page locations (as appropriate) can only have published versions (remove Draftable I think).

PR #325, opened 2020-08-13
Add content type to data payload

tomtollafield:


DFC.ServiceTaxonomy.Events/Models/ContentEvent.cs
```diff
@@ -31,7 +31,7 @@ public class ContentEvent
             string itemId = userId.Substring(userId.Length - 36);
             Subject = $"/content/{contentItem.ContentType.ToLower()}/{itemId}";
 
-            Data = new ContentEventData(userId, itemId, contentItem.ContentItemVersionId, contentItem.DisplayText, contentItem.Author, Activity.Current ?? new Activity(nameof(ContentEvent)).Start());
+            Data = new ContentEventData(userId, itemId, contentItem.ContentItemVersionId, contentItem.DisplayText, contentItem.Author, contentItem.ContentType, Activity.Current ?? new Activity(nameof(ContentEvent)).Start());
```
lazcool, line 34: Can we now remove `ContentType` from `ContentEvent`?

tomtollafield: Yeah we can, I think it's an error that it's there anyway as it doesn't get relayed from Event Grid

PR #328, opened 2020-08-17
DFCC-840 Atomic sync

lazcool:
Makes graph syncs atomic.
Also:
* fixes CypherToContent imports
* converts event project into a reusable nuget package
* only publishes event grid events if sync works
* refactor sync handling to use orchestrators


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Helpers/EmbeddedContentItemsGraphSyncer.cs
```diff
@@ -127,6 +132,13 @@ public abstract class EmbeddedContentItemsGraphSyncer : IEmbeddedContentItemsGra
             }
         }
 
+        private IMergeGraphSyncer GetNewMergeGraphSyncer()
+        {
+            _logger.LogDebug("Getting new IMergeGraphSyncer.");
```
tomtollafield, line 137: Do you still need your debug log messages?

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Helpers/EmbeddedContentItemsGraphSyncer.cs
```diff
@@ -52,6 +55,8 @@ public abstract class EmbeddedContentItemsGraphSyncer : IEmbeddedContentItemsGra
             IGraphMergeContext context,
             IAllowSyncResult allowSyncResult)
         {
+            _logger.LogDebug("Do embedded items allow sync?");
```
tomtollafield, line 58: Is this a rhetorical question haha?

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/MergeGraphSyncer.cs
```diff
@@ -102,11 +107,11 @@ public class MergeGraphSyncer : IMergeGraphSyncer
             IContentManager contentManager,
             IGraphMergeContext? parentGraphMergeContext = null)
         {
+            _logger.LogDebug("SyncAllowed to {GraphReplicaSetName} for '{ContentItem}' {ContentType}?",
```
tomtollafield, line 110: Another LogDebug here, I'm guessing it's too chatty to turn into a LogInformation and if you're confident it works now, may not be required.

tomtollafield (approved):

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Helpers/EmbeddedContentItemsGraphSyncer.cs
lazcool: They were added to help with making structural changes to the sync code, that because of recursion, callbacks and all the moving parts, makes it too cumbersome to step through it all.

You're right in that they're too chatty for logging as info (although some might be worth switching), but I think them worthwhile keeping to help with future work or bug fixes.

PR #334, opened 2020-08-20
Page location bug fixes

sturatcliffe:


DFC.ServiceTaxonomy.Taxonomies/Helper/TaxonomyHelper.cs
```diff
@@ -3,16 +3,24 @@
...
-            return contentItem.Content?[nameof(TaxonomyPart)]?[nameof(TaxonomyPart.Terms)]?.ToObject<List<dynamic>?>() ??
-                   contentItem.Content?[nameof(TaxonomyPart.Terms)]?.ToObject<List<dynamic>?>() ??
-                   contentItem[nameof(TaxonomyPart.Terms)]?.ToObject<List<dynamic>?>();
+            try
+            {
+                return contentItem.Content?[nameof(TaxonomyPart)]?[nameof(TaxonomyPart.Terms)]?.ToObject<List<dynamic>?>() ??
+                       contentItem.Content?[nameof(TaxonomyPart.Terms)]?.ToObject<List<dynamic>?>() ??
+                       contentItem[nameof(TaxonomyPart.Terms)]?.ToObject<List<dynamic>?>();
+            }
+            catch (RuntimeBinderException)
```
lazcool, line 20: Not the greatest way of handling it. Would be better to check context without an exception being thrown (using `ContainsKey`?).

lazcool (approved):

DFC.ServiceTaxonomy.Taxonomies/Helper/TaxonomyHelper.cs
sturatcliffe: Not possible on a dynamic as far as I'm aware! Tried casting to a JObject but that also blew up.

PR #340, opened 2020-08-24
Clone page handler

sturatcliffe:


DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationPartHandler.cs
```diff
@@ -1,16 +1,89 @@
...
+                newFullUrl = $"{fullUrl.Substring(0, fullUrl.IndexOf("-clone"))}{urlPostfix}";
+            }
+            else
+            {        
+                newUrlName = $"{urlName}{urlPostfix}";
+                newFullUrl = $"{fullUrl}{urlPostfix}";
+            }
+
+            context.CloneContentItem.DisplayText = title.StartsWith("CLONE - ") ? title : $"CLONE - {title}";
+            context.CloneContentItem.Alter<TitlePart>(part => part.Title = title.StartsWith("CLONE - ") ? title : $"CLONE - {title}");
```
lazcool, line 79: It only needs to generate the title/display text once.
Perhaps we should insert the already calculated clone number in the title if this is not the first clone.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationPartHandler.cs
```diff
@@ -1,16 +1,89 @@
...
+
+        public override async Task CloningAsync(CloneContentContext context, PageLocationPart part)
+        {
+            string title = context.CloneContentItem.Content[nameof(TitlePart)][nameof(TitlePart.Title)];
+            string urlName = context.CloneContentItem.Content[nameof(PageLocationPart)][nameof(PageLocationPart.UrlName)];
+            string fullUrl = context.CloneContentItem.Content[nameof(PageLocationPart)][nameof(PageLocationPart.FullUrl)];
+
+            IEnumerable<ContentItem> pages = await _session.Query<ContentItem, ContentItemIndex>(x => x.ContentType == "Page" && x.Latest).ListAsync();
+
+            string urlSearchFragment = fullUrl.EndsWith("-clone") ?
```
lazcool, line 37: I think we should have an unit test to test the url & title generation.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationPartHandler.cs
```diff
@@ -1,16 +1,56 @@
...
+
         public override Task InitializingAsync(InitializingContentContext context, PageLocationPart part)
         {
             part.Apply();
             return Task.CompletedTask;
         }
+
+        public override async Task CloningAsync(CloneContentContext context, PageLocationPart part)
+        {
+            string title = context.CloneContentItem.Content[nameof(TitlePart)][nameof(TitlePart.Title)];
```
lazcool, line 33: This logic should be moved into `MutateOnClone` in `PageLocationPartGraphSyncer`.
Apart from it belonging there, if it stayed in it's own `ContentParthandler`, we'd be dependent on the order that the 2 content handlers that handle cloning get called in, for any mutations to the content item done here, to be synced to the graph(s).

sturatcliffe: Done!

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationPartHandler.cs
sturatcliffe: Done!

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationPartHandler.cs
sturatcliffe: Done!

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/TitlePartGraphSyncer.cs
```diff
@@ -34,5 +43,23 @@ public override Task AddSyncComponents(JObject content, IGraphMergeContext conte
                 NodeTitlePropertyName,
                 context.NodeWithOutgoingRelationships.SourceNode));
         }
+
+        //we're now doing this for all cloned content items, we happy with that?
```
lazcool, line 47: We _could_ limit it to pages, but it probably makes sense for all content types.
Whatcha think?

DFC.ServiceTaxonomy.GraphSync/Services/TitlePartCloneGenerator.cs
```diff
@@ -0,0 +1,12 @@
+﻿using DFC.ServiceTaxonomy.GraphSync.Services.Interface;
+
+namespace DFC.ServiceTaxonomy.GraphSync.Services
+{
+    public class TitlePartCloneGenerator : ITitlePartCloneGenerator
+    {
+        public string Generate(string title)
+        {
+            return title.StartsWith("CLONE - ") ? title : $"CLONE - {title}";
```
lazcool, line 9: We should probably do something similar with the title as we do with the urls, so start adding a number on further clones.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/TitlePartGraphSyncer.cs
```diff
@@ -34,5 +43,23 @@ public override Task AddSyncComponents(JObject content, IGraphMergeContext conte
...
+        public override Task MutateOnClone(JObject content, ICloneContext context)
+        {
+            string? title = (string?)content[nameof(TitlePart.Title)];
+
+            if (string.IsNullOrWhiteSpace(title))
+            {
+                throw new InvalidOperationException($"Cannot mutate {nameof(TitlePart)} if {nameof(TitlePart.Title)} is missing.");
```
lazcool, line 54: If we allow a content item to have a blank title, we should allow users to clone items with blank titles.
We should just let `_titlePartCloneGenerator.Generate` handle it as normal, so it generates 'CLONE - '.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationPartHandler.cs
```diff
@@ -1,16 +1,56 @@
...
+            _generator = generator;
+        }
+
         public override Task InitializingAsync(InitializingContentContext context, PageLocationPart part)
         {
             part.Apply();
             return Task.CompletedTask;
         }
+
+        //public override async Task CloningAsync(CloneContentContext context, PageLocationPart part)
```
lazcool, line 31: Might as well delete this now - it's in git's history if we need to resurrect it.

DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Parts/TitlePartGraphSyncer.cs
sturatcliffe: Yeah all content types is fine by me!

lazcool (approved):

PR #342, opened 2020-08-24
Dfcc 936 split out service taxonomy recipe

SimonWren:


lazcool:
We should update the importer to generate a master without the extracted pages types, and leave a job profile master recipe in the repo.

lazcool (approved):

PR #347, opened 2020-08-26
Regenerate page URL's when updating page locations

sturatcliffe:


DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+                page.Alter<PageLocationPart>(part => part.FullUrl = fullUrl);
+
+                if (page.Published)
+                {
+                    page.Published = false;
+                    await _contentManager.PublishAsync(page);
+                    return;
+                }
+
+                await _contentManager.SaveDraftAsync(page);
```
lazcool, line 54: I think we need to update draft and published pages separately if necessary (including updating both graphs and publishing event grid events), or at least give this some more thought, as I don't think its as simple as re-saving draft pages.

E.g. if an item has a published version and a draft version, this will update the draft page, but the published version of the page will have an url with the old page location, and page locations are published only, so there'd be a mismatch between the published page url and the published locations.

DFC.ServiceTaxonomy.Taxonomies/Views/Admin/Edit.cshtml
```diff
@@ -19,3 +19,36 @@
     <input type="hidden" name="TaxonomyContentItemId" value="@Model.TaxonomyContentItemId" />
     <input type="hidden" name="TaxonomyItemId" value="@Model.TaxonomyItemId" />
 </form>
+
+@if (contentTypeDefinition?.DisplayName == "Page Location")
+{
+    <script depends-on="jQuery" at="Foot">
```
lazcool, line 25: Don't think it'd be too much effort to move this into the PageLocation module and add the script from the PageLocation part's view. We should keep the taxonomies module generic.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
sturatcliffe: I believe the logic in there now handles things appropriately, and the draft/pub graphs seem to keep up to date with the right state

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                //load and alter the published page
+                ContentItem publishedPage = await _contentManager.GetAsync(page.ContentItemId, VersionOptions.Published);
+
+                if (publishedPage != null)
+                {
+                    publishedPage.Alter<PageLocationPart>(part => part.FullUrl = fullUrl);
+                }
+
+                if (publishedPage != null && ((draftPage?.Latest ?? false) || draftPage == null))
```
lazcool, line 61: Don't think the `|| draftPage == null` is required.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+                    publishedPage.Published = false;
+                    await _contentManager.PublishAsync(publishedPage);
+                }
+
+                if (draftPage != null)
+                {
+                    await _contentManager.SaveDraftAsync(draftPage);
+                }
+
+                if (publishedPage != null && draftPage != null && !draftPage.Latest)
```
lazcool, line 72: Can you explain this logic to me please?
Isn't the draft page always latest, meaning this is never true?
And if `GetAsync(... VersionOptions.Draft)` returns old draft versions (not latest), then we shouldn't be saving new drafts based on a previous draft version.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                if (publishedPage != null && ((draftPage?.Latest ?? false) || draftPage == null))
+                {
+                    publishedPage.Published = false;
+                    await _contentManager.PublishAsync(publishedPage);
+                }
+
+                if (draftPage != null)
+                {
+                    await _contentManager.SaveDraftAsync(draftPage);
```
lazcool, line 69: Updating this was (republish then re save draft), will leave a small window where the preview graph contains the incorrect version (the published version, when there's a later draft version).
And if the publish succeeds and the save draft fails, the preview graph will be left out of sync with OC's draft version.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
```diff
@@ -0,0 +1,98 @@
...
+                }
+
+                if (draftPage != null)
+                {
+                    _session.Save(draftPage);
+                }
+
+                if (publishedPage != null && draftPage != null)
+                {
+                    //TODO (CHECK) : will this also fire events to the event store?
```
lazcool, line 75: No, the orchestrator doesn't fire events (it possibly _could_).
So event firing will need to be added.

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
lazcool: Can you please talk me though it, as I don't understand how it handles published and draft items correctly?

It fetches only the latest pages, so if a page has both a published and a draft version, only the the draft page is returned from the query fetching the latest pages, so surely the published version of the page doesn't get updated?

DFC.ServiceTaxonomy.GraphSync/Handlers/Orchestrators/DeleteOrchestrator.cs
```diff
@@ -45,10 +54,19 @@ public async Task<bool> Unpublish(ContentItem contentItem)
...
                 contentItem,
                 _publishedContentItemVersion,
-                DeleteOperation.Unpublish);
+                DeleteOperation.Unpublish))
+            {
+                return false;
+            }
+
+            //if unpublish was successful publish events
+            await PublishContentEvent(contentItem, ContentEventType.Unpublished);
```
lazcool, line 66: The orchestrators need to call the IContentOrchestrationHandlers, instead of publishing the events directly.

So in effect, these code snippets from `GraphSyncContentHandler`, need to be removed from there and moved down in to the orchestrator....

```
            foreach (var contentOrchestrationHandler in _contentOrchestrationHandlers)
            {
                await contentOrchestrationHandler.DraftSaved(context.ContentItem);
            }
```

DFC.ServiceTaxonomy.GraphSync/Handlers/Orchestrators/SyncOrchestrator.cs
```diff
@@ -26,16 +31,22 @@ public class SyncOrchestrator : Orchestrator, ISyncOrchestrator
...
+            IPreviewContentItemVersion previewContentItemVersion,
+            INeutralEventContentItemVersion neutralEventContentItemVersion)
+            : base(contentDefinitionManager, notifier, logger, eventGridConfiguration, eventGridContentClient, syncNameProvider, publishedContentItemVersion, previewContentItemVersion, neutralEventContentItemVersion)
         {
             _graphCluster = graphCluster;
             _publishedContentItemVersion = publishedContentItemVersion;
             _serviceProvider = serviceProvider;
         }
 
+        //Doesn't this and the other similar methods on these orchestrators return false if it failed or was blocked?
```
lazcool, line 49: You're right, would you mind changing the comments?

sturatcliffe: Done!

DFC.ServiceTaxonomy.GraphSync/Handlers/Orchestrators/DeleteOrchestrator.cs
sturatcliffe: Done!

lazcool (approved):

PR #354, opened 2020-08-27
removed SharedContent for "Send us a letter" and created ContactUsRecipe containing a single page new "Send us a letter"

IanCrisp:


GetJobProfiles/GetJobProfiles.csproj
```diff
@@ -29,6 +29,7 @@
     <Content Include="ContentRecipes\*.recipe.json">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
+    <None Remove="ContentRecipes\ContactUsPages.recipe.json" />
```
lazcool, line 32: This shouldn't be required.

IanCrisp: removed from project file

GetJobProfiles/Program.cs
```diff
@@ -195,6 +195,7 @@ static async Task Main(string[] args)
             //    await CopyRecipe(contentRecipesPath, "SharedContent");
             //    await CopyRecipe(contentRecipesPath, "ContentHelp");
                   await CopyRecipe(contentRecipesPath, "EmailTemplates");
+                  await CopyRecipe(contentRecipesPath, "ContactUsPages");
```
lazcool, line 198: If ContactUsPages is part of the pages release, it should be added to the `master_pages_release_39f603d6-746a-4324-b59a-31924e5d44fb.recipe.json` recipe instead of being part of the import util, and `ContactUsPages.recipe.json` placed in the `DFC.ServiceTaxonomy.Editor/Recipes` folder.

lazcool (approved):
ContactUsPages is not part of the initial release, so this is good to go.

PR #367, opened 2020-09-02
Check all terms down the tree when validating page location reshuffles

sturatcliffe:


DFC.ServiceTaxonomy.PageLocation/Validators/PageLocationsTaxonomyValidator.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            if (terms != null)
+            {
+                foreach (JObject term in terms)
+                {
+                    dynamic? originalParent = _taxonomyHelper.FindParentTaxonomyTerm(term, JObject.FromObject(part.ContentItem));
+                    dynamic? newParent = _taxonomyHelper.FindParentTaxonomyTerm(term, JObject.FromObject(part));
+
+                    if (originalParent == null || newParent == null)
+                        throw new InvalidOperationException($"Could not find parent taxonomy term for {(originalParent == null ? originalParent : newParent)}");
```
lazcool, line 45: Is that ternary operator the right way round?

lazcool:
I think it'd be better to update affected pages' locations, if a location is moved in the hierarchy, in a similar manner as we currently do when a location is edited.

DFC.ServiceTaxonomy.PageLocation/Validators/PageLocationsTaxonomyValidator.cs
sturatcliffe: It's just generally wrong, supposed to be the item it's looking for the parent for

sturatcliffe: Fixed now!

DFC.ServiceTaxonomy.PageLocation/Validators/PageLocationsTaxonomyValidator.cs
```diff
@@ -0,0 +1,78 @@
...
+        public PageLocationsTaxonomyValidator(IContentItemsService contentItemsService, ITaxonomyHelper taxonomyHelper, IEnumerable<ITaxonomyTermValidator> validators)
+        {
+            _contentItemsService = contentItemsService;
+            _taxonomyHelper = taxonomyHelper;
+            _validators = validators;
+        }
+
+        public async Task<TaxonomyValidationResult> Validate(TaxonomyPart part)
+        {
+            List<string> errors = new List<string>();
```
lazcool, line 30: Think this should pass validation if it's not the page location taxonomy.
(Or taxonomy validators could declare the taxonomies they should be applied to.)

sturatcliffe: Good spot, sorted!

lazcool (approved):

PR #381, opened 2020-09-08
DFCC-1013 Enhanced error notifications

lazcool:
Notification messages are stored in a cookie by Orchard Core's `NotifyFilter`.

A domain can only store roughly ~4k's worth of cookie data, which means that if the enhanced notification message is too long, instead of the notification appearing, the page errors.

This PR limits notification message's length, to stop the page from erroring.

There will be another story to fix the max length issue.


DFC.ServiceTaxonomy.GraphSync/Orchestrators/DeleteOrchestrator.cs
```diff
@@ -77,15 +77,26 @@ public async Task<bool> Delete(ContentItem contentItem)
...
-            (SyncStatus publishedSyncStatus, IDeleteGraphSyncer? publishedDeleteGraphSyncer) = deleteGraphSyncers[0];
-            (SyncStatus previewSyncStatus, IDeleteGraphSyncer? previewDeleteGraphSyncer) = deleteGraphSyncers[1];
+            (IAllowSyncResult publishedAllowSyncResult, IDeleteGraphSyncer? publishedDeleteGraphSyncer) = deleteGraphSyncers[0];
+            (IAllowSyncResult previewAllowSyncResult, IDeleteGraphSyncer? previewDeleteGraphSyncer) = deleteGraphSyncers[1];
 
-            if (publishedSyncStatus == SyncStatus.Blocked || previewSyncStatus == SyncStatus.Blocked)
+            if (publishedAllowSyncResult.AllowSync == SyncStatus.Blocked || previewAllowSyncResult.AllowSync == SyncStatus.Blocked)
             {
+                var graphBlockers = new List<(string GraphReplicaSetName, IAllowSyncResult AllowSyncResult)>();
+                if (publishedAllowSyncResult.AllowSync == SyncStatus.Blocked)
```
tomtollafield, line 90: Just an observation and I know it's existing code, but publishedAllowSyncResult.AllowSync == SyncStatus.Blocked doesn't read quite right to me. Might be better if we had something like publishedAllowSyncResult.Status or publishedAllowSyncResult.SyncStatus. In my head AllowSync would indicate a bool

tomtollafield (approved):

DFC.ServiceTaxonomy.GraphSync/Orchestrators/DeleteOrchestrator.cs
lazcool: I'll change it in the next PR to ..`AllowSync.Result == AllowSyncResult.` as discussed.

PR #383, opened 2020-09-09
DFCC 934 - Delete page location validation + validate page URL's against other page redirect locations

sturatcliffe:


lazcool (approved):

DFC.ServiceTaxonomy.PageLocation/Migrations.cs
```diff
@@ -25,7 +26,7 @@ public int Create()
 
         public int UpdateFrom1()
```
tomtollafield, line 27: Can we have a better naming convention for migrations that describe the change? If not, when we get loads it will be messy very quickly.

tomtollafield (approved):

lazcool (approved):

PR #387, opened 2020-09-10
- Update recipes for Short Question

tomtollafield:
- Import DYSaC short questions
- Add column to spreadsheet for Trait


lazcool (approved):
Just wondering about the trade-offs between having Trait as a predefined list string field vs a picked type.
Think we'll have the same decision to make on quite a few upcoming fields!

PR #393, opened 2020-09-11
DFCC-972 - Validate / repair deleted content items

sturatcliffe:


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/ValidateAndRepairGraph.cs
```diff
@@ -228,6 +253,17 @@ public async Task<ValidateAndRepairResults> ValidateGraph(params string[] graphR
                 .ListAsync();
         }
 
+        private async Task<IEnumerable<ContentItem>> GetDeletedContentItems(
+            ContentTypeDefinition contentTypeDefinition)
+        {
+            return await _session
```
lazcool, line 259: Does this get items with previous versions, as opposed to deleted content items?
If so, won't it end up deleting all current items because they are in the graph?
Also, if so, won't it fail items with multiple archive versions multiple times?

lazcool: Also, should it check the deleted time and only validate and repair items that have been deleted since the last check?

lazcool (approved):

PR #396, opened 2020-09-15
- Added handler to facilitate Event Grid publishing for a Taxonomy Term

tomtollafield:


DFC.ServiceTaxonomy.Taxonomies/Controllers/AdminController.cs
```diff
@@ -47,7 +48,8 @@ public class AdminController : Controller
             IEnumerable<ITaxonomyTermValidator> validators,
             IEnumerable<ITaxonomyTermDeleteValidator> deleteValidators,
             IEnumerable<ITaxonomyTermHandler> handlers,
-            ITaxonomyHelper taxonomyHelper)
+            ITaxonomyHelper taxonomyHelper,
+            ITaxonomyTermPublishHandler taxonomyTermPublishHandler)
```
lazcool, line 52: Could we add a new method on `ITaxonomyTermHandler` instead?
And could we call it `Published` to indicate it's handling an event, rather than `Publish`, which sounds more like an instruction.
It should also accept an `IEnumerable<IContentOrchestrationHandler>`, in case we want to add any more handlers later.

DFC.ServiceTaxonomy.Taxonomies/Controllers/AdminController.cs
```diff
@@ -297,11 +300,10 @@ public async Task<IActionResult> EditPost(string taxonomyContentItemId, string t
...
             await _contentManager.PublishAsync(taxonomy);
 
-            //force publish the new term content item to trigger events etc
-            await _contentManager.PublishAsync(contentItem);
-
+            //Content item will get published as part of the taxonomy, handler ensure Event Grid is informed of Content Item change
             foreach (var handler in _handlers)
             {
+                await handler.PublishedAsync(contentItem);
                 await handler.UpdatedAsync(contentItem, taxonomy);
```
lazcool, line 307: We need to handle errors that happen in `UpdatedAsync` better. If it can't update pages and cancels the session, then we shouldn't call `UpdatedAsync` so that events aren't published. We'd have to check that it also cancels the term change too.
Think it might need a story!

DFC.ServiceTaxonomy.PageLocation/Handlers/PageLocationTaxonomyTermHandler.cs
```diff
@@ -22,23 +22,34 @@ public class PageLocationTaxonomyTermHandler : ITaxonomyTermHandler
...
+            {
+                await orchestrator.Published(term);
+            }
+        }
+
+        public async Task<bool> UpdatedAsync(ContentItem term, ContentItem taxonomy)
         {
             if (term.ContentType != ContentTypes.PageLocation)
-                return;
+                return false;
```
lazcool, line 52: This is a check to only apply the handler's logic for `PageLocation`'s and not all taxonomy terms.
So if this handler doesn't apply, it has been successful and should return true (so that it's a no-op for other taxonomy types, rather than an error).
We also need the same check in `PublishedAsync`, so that it only applies to `PageLocation`'s.

lazcool (approved):

PR #399, opened 2020-09-15
Feature/dfcc 1046 dy sa c job profile mappings

tomtollafield:


GetJobProfiles/Program.cs
```diff
@@ -116,8 +116,12 @@ static async Task Main(string[] args)
                 }
             };
 
+            var dysacImporter = new DysacImporter();
+            using var dysacJobProfileReader = new StreamReader(@"SeedData\dysac_job_profile_mappings.xlsx");
```
lazcool, line 120: Could probably do with being moved into a class.

lazcool (approved):

PR #466, opened 2020-11-26
Various bug fixes

sturatcliffe:
Closes:

DFCC-911
DFCC-1145
DFCC-1160


DFC.ServiceTaxonomy.Taxonomies/Validation/ITaxonomyTermUpdateValidator.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace DFC.ServiceTaxonomy.Taxonomies.Validation
+{
+    public interface ITaxonomyTermUpdateValidator
```
lazcool, line 6: It'd be good if there was a single `ITaxonomyTermValidator` that had methods for validating updates, deletes etc.
Would probably have to return a bool and the error message from the validate methods.

lazcool (approved):

PR #479, opened 2021-01-06
Skills toolkit shared content update

syatessoftware:


lazcool:
I think we should have a set of recipes per release, so that we can recreate the content for a particular release.

In that case, instead of updating the `00_Pages_SharedContent.recipe.json` recipe, I think we should add a new recipe, called `01_SkillsToolkit_SharedContent.recipe.json `. I.E. the release is the 01_SkillsToolkit release, that would contain the updated 'Support during Covid-19' shared content item, and the new 'For careers advice in other parts of the UK' item.

As the recipe will need to be able to update an existing item, we'll need to use the "content" step type, instead of "contentnocache". Full explanation for that is [here](https://github.com/SkillsFundingAgency/dfc-servicetaxonomy-editor/blob/master/Documentation/Content.md).

lazcool (approved):

PR #503, opened 2021-01-25
FT2-23 Content approval spike

lazcool:
There is a [repo ](https://github.com/jeffolmstead/Rework.ContentApproval) with a content approval module, described as 'Extends Orchard Core content management with a content approval process.' The source code is licensed using the BSD-3-Clause license, which should [allow us to incorporate the code](https://softwareengineering.stacakexchange.com/questions/121998/mit-vs-bsd-vs-dual-license#:~:text=BSD-licensed%20projects%20can%20be,clause%20licenses%20are%20essentially%20identical.) into the Stax Editor (MIT licensed).

This PR contains the module added to the solution and available in the Stax Editor (once the Rework: Content Approval feature is enabled).

The module adds a custom part to manage approvals/publishing.

The custom part replaces the Publish button on the item editor page. (There is no support for approvals as part of the content items list.)

If a user doesn’t have publish permissions, it instead allows the user to request approval for the content to be published (along with a note)…

Doing so changes the status of the item (stored within the custom part) to ‘Approval Requested’. If a user with publish permissions edits the item, they see the status 'Approval Requested', along with any note entered, and they can then…

* approve, which publishes the item
* request revision, which changes the items approval status to ‘Needs Revised’
* reject, which changes the item’s approval status to ‘Rejected’

If a user has publish permission, they always see the new control with the same options as above, which doesn’t make sense if they created the item. I.e. if a user with publish permissions creates an item, they still get the option to request revisions or reject the item.

If validation fails after selecting an option from the control, the control is not subsequently shown (presumably a bug).

Workflow events are apparently published on state changes (untested as part of the spike), which would allow workflows to perform ancillary tasks, such as sending notification emails. (We’d probably want to perform any such tasks through code, rather than using workflows, which are less flexible.)

**Elements not implemented that we might need**

No filter available on the content items list to see items in a certain state.

No meta annotation of the state of content items on the content items list.

No way to request approval from a group or users (no assignment).

No visibility of items in a certain state (no dashboard).

Only a small space to enter a note (although they could change the item’s content).

No visual diff of what’s changed in the new draft version compared to the published version.

Sending notification(s) when an item’s status changes. The in-built user has an email address we could send a notification to. We could also extend the user schema to store a slack user name (we already have the facility to publish messages to slack). 

The control doesn't meet our requirements for workflow states, forced publishing etc.

**Conclusions**

The module as-is does not meet our requirements. However it should simplify our implementation, either by using it for reference, or using it as a starting point to build on.


bmatthews-DFE:
Closing due to inactivity

PR #504, opened 2021-01-25
Updated recipes for skills toolkit and shared content

IanCrisp:


lazcool:
Would you mind moving the new versions of the shared content items into the `01_SkillsToolkit_SharedContent` recipe?

Ideally, we want to be able to recreate the content for each release. E.g. if we wanted to recreate current live, we'd install the 00_Pages recipes, for skills toolkit, we'll release the 01_SkillsToolkit recipes etc.

lazcool (approved):

PR #516, opened 2021-02-04
mac compatibility fixes

thales-man:
no pack updates just the attribute changes we discussed, I moved the file into the root of the project and renamed it 'testing visibility' as it seems more pertinent.


DFC.ServiceTaxonomy.Neo4j/TestingVisibility.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("DFC.ServiceTaxonomy.UnitTests")]
+[assembly: InternalsVisibleTo("DFC.ServiceTaxonomy.IntegrationTests")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
+
+//todo: move core of validation into a validation service in the neo project
+// but for now...
+[assembly: InternalsVisibleTo("DFC.ServiceTaxonomy.GraphSync")]
```
lazcool, line 9: Would you mind changing the name of this file, as the last `InternalsVisibleTo` is functional, rather than test related.

lazcool (approved):

PR #520, opened 2021-02-09
Update Orchard Core

lazcool:
Update Orchard Core packages to include monaco-editor fork and latest Orchard Core dev branch.

Updates relevant to the Stax Editor include:
* monaco-editor available as a resource (will be needed for the html diff tool)
* monaco-editor added as an option for html and text fields
* Dashboard (required for the dashboard feature)
* fix to background task start time (now uses local time) - could affect repair workflow start time
* fix z-index issue with modals in the trumbowyg editor (we've seen the link modal not appearing in the html editor)
* fix content list search in safari, ios and pados
* fixes and optimisations to yessql (might speed up recipe import), including a [fix for recipes with >2100 items in a step](https://github.com/OrchardCMS/OrchardCore/issues/8410)
* updated html sanitizer


andycyberian (approved):
All looks good

PR #523, opened 2021-02-11
FT2-124 Dashboard

lazcool:
Adds initial content approval dashboard part and widget.

Also updates our orchard core packages with a fix applied to OrchardCore.AdminDashboard. (Our [fork has been updated](https://github.com/SkillsFundingAgency/OrchardCore/tree/monaco_editor), and a [PR](https://github.com/OrchardCMS/OrchardCore/pull/8620) raised against upstream.)


DeveloperNotes.md
```diff
@@ -1,5 +1,7 @@
 #ToDo
 
+* make open site open site (for env)
```
andycyberian, line 3: Not 100% sure what this means.

andycyberian (commented):
All looks good, just the one comment which may just be me not understanding.

DeveloperNotes.md
lazcool: There's a visit site button..
image

We should make it open the current environment's National Careers site.

andycyberian (approved):
All looks good. I see that approval statuses has been expanded on which may affect some work on content approval but we can discuss this in the refinement session.

PR #527, opened 2021-02-15
FT2-28 editing content

andycyberian:


lazcool:
The `appsettings.template.json` file has been deleted, and it's needed - one to watch out for when you delete the `app_data` folder!

DFC.ServiceTaxonomy.ContentApproval/Controllers/ContentApprovalController.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+
+            if (!await _authorizationService.AuthorizeAsync(User, Permissions.RequestReviewPermissions.RequestReviewPermission, contentItem))
+            {
+                return Forbid();
+            }
+
+            contentItem.Alter<ContentApprovalPart>(p => p.ApprovalStatus = ContentApprovalStatus.ReadyForReview);
+            await _contentManager.SaveDraftAsync(contentItem);
+            _notifier.Success(H[$"{contentItem.DisplayText} is now ready to be reviewed."]);
```
lazcool, line 41: The string needs to be parameterized using `{0}`, as otherwise the content item's display text will be used when looking up the appropriate string to translate, and it won't be found.

DFC.ServiceTaxonomy.ContentApproval/Controllers/ContentApprovalController.cs
```diff
@@ -0,0 +1,46 @@
...
+using OrchardCore.DisplayManagement.Notify;
+
+namespace DFC.ServiceTaxonomy.ContentApproval.Controllers
+{
+    public class ContentApprovalController : Controller
+    {
+        private readonly IAuthorizationService _authorizationService;
+        private readonly IContentManager _contentManager;
+        private readonly INotifier _notifier;
+        private readonly IHtmlLocalizer<ContentApprovalController> H;
```
lazcool, line 16: I think `IStringLocalizer` should be sufficient here.

DFC.ServiceTaxonomy.ContentApproval/DFC.ServiceTaxonomy.ContentApproval.csproj
```diff
@@ -29,4 +30,14 @@
     </PackageReference>
   </ItemGroup>
 
+  <ItemGroup>
```
lazcool, line 33: I don't _think_ this change is needed.

DFC.ServiceTaxonomy.ContentApproval/Shapes/UserEditShapes.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        public void Discover(ShapeTableBuilder builder)
+        {
+            builder.Describe("Content_SummaryAdmin").OnDisplaying(context =>
+            {
+                var currentUser = _httpContextAccessor.HttpContext?.User;
+                if (currentUser == null || !currentUser.IsInRole("Editor"))
```
lazcool, line 23: Feels like we should be checking a permission here, rather than a role?

DFC.ServiceTaxonomy.ContentApproval/Controllers/ContentApprovalController.cs
andycyberian: The notifier looks to require an IHtmlLocalizer response as an input parameter.

lazcool (approved):

PR #555, opened 2021-03-08
FT2-212 + FT2-213 dashboard drag and drop and resize

lazcool:
Makes the dashboard edit controls visible in edit mode on the content approval status widgets.

Note: there seems to be an issue in OC's drag and drop implementation, where the dragged widget expands/contracts when being dragged.

Also:
* adds the 'New' admin menu when setting up a fresh site (whether it appears is now configurable, and switched off by default)


andycyberian (approved):
All looks good

PR #563, opened 2021-03-11
FT2-289 Update config to match with devops pipeline config

pramireddy:
Updated config file to match with DevOps pipeline config.


andycyberian (approved):
All looks good

PR #588, opened 2021-03-18
FT2-310 Added code for Youtube url query params

pramireddy:
Added code to set query params for Youtube Link URL.
Updated Inert YoutuneLink Modal dialog to set Title as required


andycyberian (approved):
Looks good to me

PR #595, opened 2021-03-24
FT2-258 ensure dropping index table that is not there doesn't cause a…

andycyberian:
…n issue


lazcool (approved):

DFC.ServiceTaxonomy.ContentApproval/Migrations.cs
```diff
@@ -34,7 +34,15 @@ public async Task<int> CreateAsync()
...
 
-            SchemaBuilder.DropMapIndexTable<ContentApprovalPartIndex>();
+
+            try
+            {
+                SchemaBuilder.DropMapIndexTable<ContentApprovalPartIndex>();
+            }
+            catch
+            {
+                // Not required by SQLLite as no issue if index doesn't exist but maybe a problem in SQL
```
lazcool, line 44: Can we add logging in the catch blocks, to see if your theory is correct?

lazcool (approved):

lazcool (approved):

PR #607, opened 2021-04-08
FT2-359 shared content issue when publishing

andycyberian:


pramireddy (approved):
Code looks fine for me

PR #609, opened 2021-04-12
FT2-360 fix to allow dashboard widget to be deleted

andycyberian:
GraphSyncPart added to content approval dashboard widget to allow it to be deleted.


DFC.ServiceTaxonomy.GraphSync/Orchestrators/DeleteOrchestrator.cs
```diff
@@ -75,6 +75,13 @@ public async Task<bool> Delete(ContentItem contentItem)
             (IAllowSync publishedAllowSync, IDeleteGraphSyncer? publishedDeleteGraphSyncer) = deleteGraphSyncers[0];
             (IAllowSync previewAllowSync, IDeleteGraphSyncer? previewDeleteGraphSyncer) = deleteGraphSyncers[1];
```
Gary-Griffiths-Methods, line 76: Do we need to consider nulls?

Gary-Griffiths-Methods (approved):

DFC.ServiceTaxonomy.GraphSync/Orchestrators/DeleteOrchestrator.cs
andycyberian: As far as I can see Orchestrator.GetDeleteGraphSyncerIfDeleteAllowed(...) will throw an exception if it fails so I don't see this being a problem.

Gary-Griffiths-Methods:
Good stuff, it looks like there’s a lot to take into consideration when deleting graph items!

PR #610, opened 2021-04-14
FT2-360 avoid event grid if no graphsyncpart present

andycyberian:
If there is no GraphSyncPart there is no need to update Event Grid as the content item won't be sync'd to Neo4j

### DIFF
--- a/DFC.ServiceTaxonomy.GraphSync/Handlers/EventGridPublishingHandler.cs
+++ b/DFC.ServiceTaxonomy.GraphSync/Handlers/EventGridPublishingHandler.cs
@@ -121,6 +121,12 @@ namespace DFC.ServiceTaxonomy.GraphSync.Handlers
                 return;
             }
 
+            if (context.ContentItem.Content.GraphSyncPart == null)
+            {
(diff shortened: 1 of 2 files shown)


DFC.ServiceTaxonomy.GraphSync/Handlers/EventGridPublishingHandler.cs
```diff
@@ -121,6 +121,12 @@ public async Task Restored(IOrchestrationContext context)
                 return;
             }
 
+            if (context.ContentItem.Content.GraphSyncPart == null)
```
pramireddy, line 124: I think it's better to stop firing this event instead  of adding condition here.

andycyberian: This code is triggered by multiple events all of which would need to be modified.
Also if there was a requirement for a new event to call this code the dev would need to know to test for this condition which is something that could be missed.

pramireddy (approved):
Changes are looks fine for me.

PR #621, opened 2021-04-27
Ft2 351 careersadvice migration

pramireddy:
Added recipe files for Careers Advice static content
Update styling to use nationalcareers_toolkit


Gary-Griffiths-Methods (approved):
As these are recipe files I can't run them locally to confirm they work without the risk of breaking my own environment which I'm using for the JobProfiles work at the moment so have approved them on the basis that visually they look okay.

PR #636, opened 2021-05-19
Added new content type for Triage tool, with their initial baseline c…

IanCrisp:
…ontent

Updated Page content type to include new triage tool fields and new Triage Tool tab
Updated Placements for new triage tool tab


andycyberian (approved):
All looks good me, approved.

PR #643, opened 2021-05-24
Fix for missing graph sync on text area widget

IanCrisp:
Removal of content approvals from Home page content items


andycyberian (approved):
All looks good Ian

PR #653, opened 2021-05-27
ft2-560 fixing placements

andycyberian:


pramireddy (approved):
Chanages are looks fine.

PR #659, opened 2021-06-07
Release 1.1.5.1

andycyberian:
NCS-91 - Stax editor dashboard
NCS-95 - Ability to add images and video to content and upload to CDN
NCS-98 - Implement approval and publishing workflow in line with the content strategy

PR #661, opened 2021-06-07
Revert "Release 1.1.5.1"

andycyberian:
Reverts SkillsFundingAgency/dfc-servicetaxonomy-editor#659


Gary-Griffiths-Methods (approved):
There's a lot of changes here, and without knowledge of what was done it's difficult to comment so I'm approved it as it will be quicker to see if it fails the integration test during the build.

PR #678, opened 2021-06-15
Ft2 611 oc upgrade fix

andycyberian:


pramireddy (approved):
Code chages are looks fine.

PR #718, opened 2021-07-30
Created new TT recipe for PP import

IanCrisp:


ryanbarlowmethods (approved):
Checked with Ian there are no concept types that didnt yet want to be promoted so we dont get the same issues as previously - all good AFAIK so approving

PR #732, opened 2021-08-24
FT2-893 Review custom event not being logged. Added code for the 'Rev…

Gary-Griffiths-Methods:
…iew' audit trail event.

### DIFF
--- a/DFC.ServiceTaxonomy.ContentApproval/Handlers/ContentApprovalAuditTrailEventHandler.cs
+++ b/DFC.ServiceTaxonomy.ContentApproval/Handlers/ContentApprovalAuditTrailEventHandler.cs
@@ -94,6 +94,7 @@ namespace DFC.ServiceTaxonomy.ContentApproval.Handlers
                             // Request Review/Save Draft/Save Draft & Exit button
                             else if (button.Contains(Constants.SubmitSaveKey))
                             {
+                                string reviewType = string.Empty;
                                 switch (action)
                                 {
                                     case "submit.Save":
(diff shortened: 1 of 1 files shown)


DFC.ServiceTaxonomy.ContentApproval/Handlers/ContentApprovalAuditTrailEventHandler.cs
```diff
@@ -105,19 +106,31 @@ public override Task CreateAsync(AuditTrailCreateContext context)
                                         break;
 
                                     case "submit.RequestApproval-ContentDesign":
-                                        contentEvent.Name = "Requested for review (ContentDesign)";
+                                        reviewType = "content design";
+                                        contentEvent.Name = contentApprovalPart.ReviewStatus != Models.Enums.ReviewStatus.InReview
```
ravi-methods, line 110: Could we extract these 3 lines to a local method and call it SetContentEventName or something? This is being used in 4 places.

ravi-methods: Or rather GetContentEventName, so you do the assignment here by just getting the value from that method.

Gary-Griffiths-Methods: Added GetContentEventName() method

DFC.ServiceTaxonomy.ContentApproval/Handlers/ContentApprovalAuditTrailEventHandler.cs
```diff
@@ -136,5 +136,13 @@ public override Task CreateAsync(AuditTrailCreateContext context)
             return Task.CompletedTask;
         }
 
+        private static string GetContentEventName(AuditTrailCreateContext<AuditTrailContentEvent> contentEvent, ContentApprovalPart contentApprovalPart, string reviewType)
+        {
+            contentEvent.Name = contentApprovalPart.ReviewStatus != Models.Enums.ReviewStatus.InReview
```
ravi-methods, line 141: Don't need to pass contentEvent, just return the string value itself without assigning it to anything.

DFC.ServiceTaxonomy.ContentApproval/Handlers/ContentApprovalAuditTrailEventHandler.cs
Gary-Griffiths-Methods: Removed ContentEvent parameter.

ravi-methods (approved):

PR #752, opened 2021-09-15
FT2-912-Custom-events-force-publish-creates-2-identical-audit-events. The two identical events are due to a 'save draft'  being done. Have …

Gary-Griffiths-Methods:
…added code to write the second event correctly.


pramireddy (approved):
Code changes are looks fine.

PR #757, opened 2021-09-27
Add code to fix bug in BannerContentPicker

pramireddy:
MInor code change to fix bug in Banner Content Picker. Banner content picker works only if  dropdown list itemtype is Banner.


andycyberian (approved):
Approved

PR #779, opened 2021-11-19
FT2-1020 College entry requirements

AnatoliMalev-Methods:


DFC.ServiceTaxonomy.Editor/Recipes/245.JobProfileCollegeentryrequirements.recipe.json
```diff
@@ -0,0 +1,158 @@
+{
+  "name": "JobProfileCollegeentryrequirements",
```
ravi-methods, line 2: Is there an issue if we make it PascalCase like JobProfileCollegeEntryRequirements for every occurance of this word within this file?

ravi-methods (approved):

PR #859, opened 2022-01-06
Ft2 1143

manluck1:


DFC.ServiceTaxonomy.Editor/App_Data/Sites/Default/appsettings.template.json
```diff
@@ -1,5 +0,0 @@
-﻿{
```
ravi-methods, line 1: Can't delete this, it's used by the pipeline to populate connection strings during deployment.

PR #861, opened 2022-01-06
Ft2 1143

manluck1:


.editorconfig
```diff
@@ -107,7 +107,7 @@ dotnet_style_prefer_conditional_expression_over_return = true:refactoring
 csharp_prefer_simple_default_expression = true:suggestion
 
 # Expression-bodied members
-csharp_style_expression_bodied_methods = true:refactoring
```
ravi-methods, line 110: I'd leave it to refactoring rather than none

ravi-methods: or rather revert it back to what it was.

PR #865, opened 2022-01-07
Ft2 1143

manluck1:


DFC.ServiceTaxonomy.JobProfiles.Module/ServiceBusHandling/DataEventProcessor.cs
```diff
@@ -114,17 +119,41 @@ private async Task GenerateServiceBusMessageForJobProfile(ContentContextBase con
...
+            {
+                if (isSaved)
                 {
-                    new WhatYouWillDoContentItem()
-                    {
-                        Id = context.ContentItem.As<GraphSyncPart>().ExtractGuid(),
-                        Title = context.ContentItem.Content.Title}
-                };
-            return _serviceBusMessageProcessor.SendOtherRelatedTypeMessages(jobprofileData, context.ContentItem.ContentType, actionType);
+                    jobprofileData.Add(new WhatYouWillDoContentItem() { Id = context.ContentItem.As<GraphSyncPart>().ExtractGuid(), Title = context.ContentItem.Content.TitlePart.Title,
```
robmuk0, line 142: The formatting is a bit off making it difficult to read, please format the document (Hint: Ctrl+K, Ctrl+D) :)

andycyberian (approved):

DFC.ServiceTaxonomy.JobProfiles.Module/Repositories/IJobProfileRepository.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Linq.Expressions;
+using YesSql;
+using YesSql.Indexes;
+
+namespace DFC.ServiceTaxonomy.JobProfiles.Module.Repositories
+{
+    public interface IJobProfileRepository<T> : IDisposable where T : MapIndex
```
ravi-methods, line 8: The implementation suggests it can be any Index because it's generic and can be used with any index by the consumer, so would it be worth changing it to something like IRepository<T> and Repository<T>? Could also be useful in other projects within STAX, could even create a separate class lib called DFC.ServiceTaxonomy.Common and move it there perhaps?

DFC.ServiceTaxonomy.JobProfiles.Module/Migrations.cs
```diff
@@ -71,5 +71,27 @@ public int Create()
...
         }
+
+        public int UpdateFrom1()
+        {
+            _contentDefinitionManager.AlterPartDefinition(nameof(JobProfile), part => part
+                .Attachable()
+                .WithDescription("Creates JobProfile related metadata record.")
+            );
+
+            SchemaBuilder.AlterIndexTable<JobProfileIndex>(table => table.AddColumn<string>(nameof(JobProfileIndex.JobProfileTitle)));
```
ravi-methods, line 82: Can this column be added to Create method instead as this code hasn't been deployed anywhere yet, so a single create method should be okay instead of a create and an update...? Could be an issue for you locally though.

PR #904, opened 2022-02-01
Ft2 1119

manluck1:


DFC.ServiceTaxonomy.JobProfiles.Module/ServiceBusHandling/Converters/Helper.cs
```diff
@@ -108,5 +108,15 @@ contentItem.ContentType switch
                ContentTypes.ApprenticeshipStandard => contentItem.Content.ApprenticeshipStandard.LARScode.Text,
                _ => string.Empty,
            };
+
+        public static string GetHtmlWithoutInlineCss(dynamic html)
```
ravi-methods, line 112: The method name says without InlineCss but it's removing all html tags? Could perhaps use a better name? and is this code ported from DFC.Digital?

ravi-methods (approved):

PR #909, opened 2022-02-03
FT2-1202

manluck1:


DFC.ServiceTaxonomy.JobProfiles.Module/ServiceBusHandling/Converters/Helper.cs
```diff
@@ -74,14 +75,26 @@ public static async Task<IEnumerable<string>> GetRelatedSkillsAsync(dynamic cont
             return Enumerable.Empty<string>();
         }
 
-        private static string GetSlugValue(string field)
+        public static string GetSlugValue(string field, bool isForwardSlashNeeded = default, bool isPascalCaseUrl = default, bool isOnlyHyphenated = default)
```
ravi-methods, line 78: This is too many responsibilities for a single method. it feels counter intuitive. I'd rather keep it simple and have a method for each of the separate outcome. Like Slugify, Hyphenate etc.

ravi-methods (changes requested):
Might want to revert back, or simplify.

DFC.ServiceTaxonomy.JobProfiles.Module/ServiceBusHandling/Converters/Helper.cs
manluck1: It handles 3 conditions. I can create 3 + 1 methods but there will be some code repetition, all doing the same thing with a slight difference.

andycyberian: Could you combine these into some kind of 'SlugOptions' class parameter?

ravi-methods: Urls are case insensitive, which is why I don't think isPascalCaseUrl is required. and isOnlyHyphenated is returning the result straight away, so wherever consumer needs a slugify and not just hyphenate, would need to go through this condition unnecessarily. It's increasing the cyclomatic complexity of the method. So if you consider that, then all you need are 2 methods, Slugify and Hyphenate.

DFC.ServiceTaxonomy.JobProfiles.Module/ServiceBusHandling/Converters/Helper.cs
```diff
@@ -74,13 +75,17 @@ public static async Task<IEnumerable<string>> GetRelatedSkillsAsync(dynamic cont
             return Enumerable.Empty<string>();
         }
 
-        private static string GetSlugValue(string field)
+        public static string GetSlugValue(string field)
```
ravi-methods, line 78: Could you please move these 2 methods to StringExtensions class as well?

PR #919, opened 2022-02-14
FT2-1257

manluck1:


DFC.ServiceTaxonomy.JobProfiles.Module/ServiceBusHandling/Converters/Helper.cs
```diff
@@ -107,13 +107,13 @@ contentItem.ContentType switch
         public static string SanitiseHtml(dynamic html)
         {
             string htmlString = ((string)html).Replace("<br>", "");
-            return Regex.Replace(htmlString, "<p[^>]*>|</p[^>]*>|</span[^>]*>|<span[^>]*>|^<ul><li>|</li></ul>$", "");
+            return Regex.Replace(Regex.Replace(htmlString, "<p[^>]*>|</p[^>]*>|</span[^>]*>|<span[^>]*>|^<ul><li>", ""), "</li></ul></li></ul>$", "</li></ul>");
```
ravi-methods, line 110: Not sure what it's going to fix? the issue we have is that the last element before "Entry requirement" isn't wrapped in a p tag, this doesn't look to be fixing that.

manluck1:
This is not needed now as Sanitisers are not required

PR #959, opened 2022-04-06
FT2-1379 cosmos service refactor

andycyberian:


DFC.ServiceTaxonomy.GraphSync/CosmosDb/CosmosDbService.cs
```diff
@@ -78,5 +97,32 @@ public async Task DeleteItemAsync(Container container, string contentType, Guid
             var result = await iteratorLoop.ReadNextAsync();
             return result.Resource.SingleOrDefault();
         }
+
+        public async Task<List<T>> GetItemQueryAsync<T>(string databaseName, string query, string contentType)
```
ravi-methods, line 101: Could perhaps name it like GetItemsAsync or GetDocumentsAsync as we're not returning Query but returning actual items.. :)

ravi-methods (approved):

PR #1023, opened 2022-06-16
FT2-1542 fixes for string decimals and other numerics

andycyberian:


DFC.ServiceTaxonomy.GraphSync/GraphSyncers/Fields/NumericFieldGraphSyncer.cs
```diff
@@ -55,8 +55,16 @@ public async Task AddSyncComponents(JObject contentItemField, IGraphMergeContext
...
+                bool longsSame;
+
+                if (!long.TryParse(nodePropertyValue.ToString(), out long nodePropertyValueLong))
+                {
+                    longsSame = false;
+                }
+                else
+                {
+                    longsSame = nodePropertyValueLong == (long)contentItemFieldValue;
+                }
```
ryanbarlowmethods, line 67: Does the double check below need to do something similar?

PR #1081, opened 2024-01-16
Feature/ad 164877

gcompton-DFE:
Added code for invalidating content in Redis.


Resources/ArmTemplates/template.json
```diff
@@ -238,6 +238,9 @@
         },
         "workspaceResourceGroup": {
             "type": "string"
+        },
+        "CmsRedisCacheConnectionString": {
+        "type": "string"
```
thelious1974, line 243: Please make this 'securestring',  so credentials aren't accidentally leaked via the Azure Portal.

https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/data-types#secure-strings-and-objects for more info.

PR #1108, opened 2024-03-19
Ad 174649

JamesPrior-07:
Changes to stax to update app registry for newly created or updated pages due to issues with cosmos removal from pages application


gcomptonedu (approved):
Reviewed.

PR #1194, opened 2024-09-16
Add functionality to send Pages, banners, page banners and shared con…

JPrior-DFE:
…tent event messages from STAX to the event grid


rpedu (approved):

rpedu (approved):

DFC.ServiceTaxonomy.CompUi/Handlers/CacheHandler.cs
```diff
@@ -232,4 +327,19 @@ private Processing GetProcessingData(ContentContextBase currentContext, string p
...
     }
+
+    //Temp method to determine if page route matches what should currently be processed.
+    //Can be removed once all routes are fine to be added.
+    private bool PageRouteFlag(Processing processing)
+    {
+        //Temp check to filter pages by certain allowed url's/titles
+        var current = JsonConvert.DeserializeObject<ContentItem>(processing.CurrentContent);
+        bool isAllowedPage = true;
+        if (processing.ContentType == "Page" && !eventGridAllowedPages.Any(pageUrl => current.PageLocationParts.FullUrl.Contains(pageUrl)))
```
abo-olsen-DFE, line 338: Instead of using a string, use the ContentTypes.Page as seen in the `switch` statement above.

```suggestion
        if (processing.ContentType == nameof(ContentTypes.Page) && !eventGridAllowedPages.Any(pageUrl => current.PageLocationParts.FullUrl.Contains(pageUrl)))
```

DFC.ServiceTaxonomy.CompUi/Handlers/CacheHandler.cs
```diff
@@ -232,4 +327,19 @@ private Processing GetProcessingData(ContentContextBase currentContext, string p
...
         return null;
     }
+
+    //Temp method to determine if page route matches what should currently be processed.
+    //Can be removed once all routes are fine to be added.
+    private bool PageRouteFlag(Processing processing)
+    {
+        //Temp check to filter pages by certain allowed url's/titles
+        var current = JsonConvert.DeserializeObject<ContentItem>(processing.CurrentContent);
+        bool isAllowedPage = true;
```
abo-olsen-DFE, line 337:
```suggestion
        var isAllowedPage = true;
```

DFC.ServiceTaxonomy.CompUi/Handlers/EventGridHandler.cs
```diff
@@ -47,7 +41,7 @@ public async Task SendEventMessageAsync(Processing processing, ContentEventType
                             await _eventGridClient.Publish(new ContentEvent(eventData, ContentEventType.StaxDelete, StaxDeleteMessage + eventData.ContentType));
                             break;
                         default:
-                            _logger.LogError($"Current content could not be retrieved for Content Item ID: {processing.ContentItemId}");
+                            _logger.LogError($"Current content could not be retrieved for Content Item ID: {contentData.ContentItemId}");
```
abo-olsen-DFE, line 44: Is this message in the wrong place? Should it have been in an `else` for the `if (eventData != null)`?

DFC.ServiceTaxonomy.CompUi/Repository/DataRepository.cs
```diff
@@ -0,0 +1,64 @@
...
+                    $"FROM Document AS D " +
+                    $"INNER JOIN ContentItemIndex AS CII on CII.DocumentId = D.Id " +
+                    $"WHERE D.Id IN ({queryIds}) " +
+                    $"AND Latest = 1 AND Published = 1";
+
+                resultList = await ExecuteQuery<RelatedContentData>(sql);
+            }
+
+            return resultList;
+        }
```
abo-olsen-DFE, line 49: This is really scary. I really hope you know where the data is coming from, as otherwise this seems like an easy target for SQL injection.

A string should never be put directly into a query, but parameterization should be used instead. Unfortunately, the wrapper that has been created does not seem to support this. I would highly encourage to get the wrapper updated to support parameterization and get this query in better shape!

https://www.learndapper.com/parameters

abo-olsen-DFE (commented):
Overall, I think it looks good. There are few small things I would change, and then there is one I think should be changed, which is the SQL parameterization. Unless you are 112% certain that data always comes from a trusted source, then it may be okay, however, I would rather be safe than sorry, and just get it fixed.

PR #1232, opened 2024-11-18
Release/system simplification/release 72.8

rpedu:
Updated preview functionality to use relative url for redirection + existing preview functionality for as-is not to use content main app call for token verification.


gcompton-DFE (approved):
Reviewed

PR #1238, opened 2024-11-26
Ssp2/ad 203602 job profile preview function

JPrior-DFE:


Resources/ArmTemplates/parameters.json
```diff
@@ -259,6 +259,18 @@
         },
         "PreviewRedirectUri:1:ContentType": {
             "value": "__ContentType1__"
+        },
+        "PreviewRedirectUri:2:RedirectUri": {
```
zlu-DFE, line 263: It makes more sense to change 2 to footer?

Resources/ArmTemplates/template.json
```diff
@@ -266,6 +266,18 @@
         },
         "PreviewRedirectUri:1:ContentType": {
             "type": "string"
+        },
+        "PreviewRedirectUri:2:RedirectUri":{
```
zlu-DFE, line 270: 2 - footer
3 - jobsector
etc. this number is not good name conversion. We need give the name more meaningful.

Resources/ArmTemplates/parameters.json
JPrior-DFE: The values have to be written like this so it can be compiled into a proper list in the appsettings and can be easily retrieved from the app without writing specific configuration retrieval for each content type

zlu-DFE (approved):

PR #1243, opened 2024-11-28
Ssp2/us 205943 update data field size

gcompton-DFE:
Updating the field size of the RelatedContentIds field in the RelatedContentItemIndex table.


DFC.ServiceTaxonomy.CompUi/Migrations.cs
```diff
@@ -41,21 +41,21 @@ public int UpdateFrom1()
...
+            );
 
-                SchemaBuilder.AlterTable(nameof(RelatedContentItemIndex), table => table
-                    .AddColumn<string>("RelatedContentIds", c => c.WithLength(1024))
-                );
+            SchemaBuilder.AlterTable(nameof(RelatedContentItemIndex), table => table
+                .AddColumn<string>("RelatedContentIds", c => c.WithLength(1024))
+            );
 
             return 3;
```
zlu-DFE, line 60: Why we have hard coded value here?

abo-olsen-DFE (approved):

PR #1262, opened 2025-01-29
Release 77.5 into MASTER

bmatthews-DFE:


rpedu (approved):
unexpected .net version been identified for three project files (triage tool migration, explore careers and homepage) but this should be the desired state hence approving the PR.

PR #1269, opened 2025-02-04
Ssp2/us 194367 send event grid messages

gcompton-DFE:


DFC.ServiceTaxonomy.CompUi/Handlers/CacheHandler.cs
```diff
@@ -250,17 +250,30 @@ private async Task ProcessRelatedContent(Processing processing)
     {
         var result = await _relatedContentItemIndexRepository.GetRelatedContentDataByContentItemIdAndPage(processing);
 
-        var contentTypes = new List<string> {
+        if (result != null)
+        {
+            if (result.ToList().Any(x => x.ContentType == nameof(ContentTypes.PersonalityShortQuestion)) && processing.ContentType == nameof(ContentTypes.PersonalityTrait))
+            {
+                var item = result.FirstOrDefault(x => x.ContentType == nameof(ContentTypes.PersonalityShortQuestion));
```
rpedu, line 257: Can you add a comment for to indicate we are restricting to many events generated for personalityshortquestion when personality trait was updated. so that meta data for shortquestion updated only once by the cache handler.

gcompton-DFE: Added comments to the relevant section of code.

PR #1277, opened 2025-02-21
Bug fix 218978 - Newly created job profiles return 404 in explore careers for around 404

mimran-DFE:
Bug fix 218978 - Newly created job profiles return 404 in explore careers for around 404


mimran-DFE:
I ran this locally and debugged the code. I found the issue while debugging the code. In the scenario of Create job profile, previousContent object is passed as null and it fails while running following code
JsonConvert.DeserializeObject<ContentItem>(previousContent);

mpasumarthi-DFE (approved):
Approved

PR #1279, opened 2025-02-24
AD-212667: STAX export of all URLs associated all job profiles

hnandra-DFE:


hnandra-DFE:
Do not close as it is used on request from the Data Insights Team

PR #1296, opened 2025-04-23
AD-211843: Remove SHC related recipes from STAX

hnandra-DFE:


pvoorukonda-DFE (approved):
looks good. removed SHC recipe files

PR #1305, opened 2025-05-27
Added recipe files for basic card, templates and placements

siva-muruganandham:


siva-muruganandham:
AllPlacement and AllTemplates are not only for FAC, They could contain the placement and templates that are used in different part of the application

